PR #18, opened 2021-07-09
Turn on cpp test to bypass abseil-py's build error

cynthiajoan:


a-maurice:
I don't think that this is a great solution, because it'll mean building a ton of tests that we don't actually want.

cynthiajoan:
> I don't think that this is a great solution, because it'll mean building a ton of tests that we don't actually want.

I agree that's not an ideal one, better solution is to use -DBUILD_TESTING=off, but unfortunately that doesn't work. And since we are not directly refer abseil-py, it's the libs like boringssl or curl did that, even the flag -DBUILD_TESTING=off is already set on cpp's CMakeList.txt, the error would still hit the second run.

chkuang-g (dismissed):

cynthiajoan:
Update the PR, only turn on stub flag for CPP. corresponding CPP PR is https://github.com/firebase/firebase-cpp-sdk/pull/528

build_bash.sh
```diff
@@ -36,6 +36,7 @@ fi
 
 CMAKE_OPTIONS="${CMAKE_OPTIONS} -DUNITY_ROOT_DIR=${UNITY_ROOT_DIR}"
 CMAKE_OPTIONS="${CMAKE_OPTIONS} -DFIREBASE_UNITY_BUILD_TESTS=ON"
+CMAKE_OPTIONS="${CMAKE_OPTIONS} -DFIREBASE_CPP_BUILD_STUB_TESTS=ON"
```
a-maurice, line 39: Could you add a comment explaining why this is needed

cynthiajoan: DONE

a-maurice (approved):

PR #29, opened 2021-07-20
Workflow install/activate/deactivate unity license

sunmou99:
Trigger on `push` is only for the test purposes. Will remove this trigger before merge. 
Manual trigger `workflow_dispatch ` won't work (cannot see “Run workflow” button on the web UI) until merge to main branch. Will try manual trigger after merge.

Unity License & Account are stored as SECRETs.
Critical steps added timeout.
Deactivate step added retry.

Workflow runs [successfully](https://github.com/firebase/firebase-unity-sdk/actions/runs/1049738500). 

Note:
Deactivate step with `if: always()` will run even cancelled, but will not run when workflow has critical errors. [source](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always).


StephenHodgson:
Why not use the Unity Hub exe to do installs?

This python script doesn't cover but a few edge cases

sunmou99:
> Why not use the Unity Hub exe to do installs?
> 
> This python script doesn't cover but a few edge cases

1) We will add support to all OSs (Windows, MacOS, Linux). Does Unity Hub provide CLI to do installs?
2) What are the edge cases?
Thank you!

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          python scripts/gha/unity_installer.py --activate_license \
+            --version ${{ matrix.unity_version }}  \
+            --username "${{ secrets.UNITY_USERNAME }}" \
+            --password "${{ secrets.UNITY_PASSWORD }}" \
+            --serial_ids "${{ secrets.SERIAL_ID }}" \
+            --logfile testapps/activate_license.log
+          cat testapps/activate_license.log
+      - name: Return Unity license
+        uses: nick-invision/retry@v2
+        if: always()
```
a-maurice, line 87: Can you add a comment about this.  Namely, what the retry thing does, and why it is using always

sunmou99: Comment added

a-maurice (dismissed):

sunmou99:
Removed push trigger.

a-maurice (approved):

StephenHodgson:
> We will add support to all OSs (Windows, MacOS, Linux). Does Unity Hub provide CLI to do installs?
yes, afaik.

> What are the edge cases?
> Thank you!

Well there's other modules to install, and the hub handles all that. but if it works as is then you guys should be fine.

PR #43, opened 2021-08-03
Remove swig update logic, add openssl, and fix iOS packaging

a-maurice:


.github/workflows/sdk_build.yml
```diff
@@ -104,6 +90,17 @@ jobs:
...
+      - name: Install OpenSSL (Linux)
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          sudo apt install openssl
+      - name: Install OpenSSL (macOS)
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          brew install openssl
+          # brew won't overwrite MacOS system default OpenSSL, so force it here.
+          echo "OPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1" >> $GITHUB_ENV
```
vimanyu, line 102: Could this hardcoded path break if openssl versions is updated in brew and is no longer found in  `openssl@1.1`?
Maybe it makes sense to use `brew --prefix openssl` to get the path?

vimanyu (commented):
One suggestion on brew installation but looks good otherwise.

vimanyu (approved):
LGTM

PR #48, opened 2021-08-12
Add a license header to all Firestore source files

var-const:


var-const:
Just for completeness' sake.

ehsannas (approved):

PR #50, opened 2021-08-17
Integration Test GHA run test workflow

sunmou99:
Enable Test running in workflow. The goal is simply to make sure the workflow works. Will have following PRs for further improvement.
Workflow run: https://github.com/firebase/firebase-unity-sdk/actions/runs/1141931498

Major changes:
Upload testapps as artifacts.
Add Test jobs for all platform.
Modify & print out test results.

TODO: 
Have a configuration file to set build & test matrix.
Fix test errors & failures. 
Speed up the build time.
Add testing on mobile simulators.
etc.


.github/workflows/integration_tests.yml
```diff
@@ -118,4 +116,139 @@ jobs:
           timeout_minutes: 10
           max_attempts: 3
           command: |
-            python scripts/gha/unity_installer.py --release_license --logfile testapps/release_license.log --version ${{ matrix.unity_version }}
+            python scripts/gha/unity_installer.py --release_license \
+              --logfile testapps/release_license.log \
+              --version ${{ matrix.unity_version }}
+      - name: Upload Android integration tests artifact
```
sunmou99, line 122: Anyone knows how to better handle the artifacts for different platforms. 
(Have a script is an option.)

yifanyang: Would this [composite action](https://github.com/actions/runner/issues/646#issuecomment-901336347) work for you?

Essentially, we can create our own template action that executes other actions such as checking out code, setting up environment, running test, and uploading artifacts, where the step of running test can be customized. Then when we use this template action in the workflow yaml file, we only need to specify the running test part, while artifact uploading is configured implicitly.

.github/workflows/integration_tests.yml
```diff
@@ -97,6 +92,7 @@ jobs:
         run: |
           python scripts/gha/restore_secrets.py --passphrase "${{ secrets.TEST_SECRET }}"
       - name: Build integration tests
+        timeout-minutes: 220
```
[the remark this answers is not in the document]
sunmou99: Yes. It can take up to 3.33 hours. Will try to speed up later

.github/workflows/integration_tests.yml
a-maurice: I assume you mean how to reduce having to duplicate it a bunch?  One potential way is to make all of the platform settings variables that get set as part of the matrix, though that can also be confusing to read.  I think starting with a readable, if a bit wordy, solution is fine for now.

scripts/gha/build_testapps.py
```diff
@@ -522,8 +532,72 @@ def run_xcodebuild(dir_helper, ios_config, device_type):
         configuration=ios_config.configuration)
 
 
-def _summarize_results(testapps, platforms, versions, failures, output_dir):
+def _collect_integration_tests(config, testapps, root_output_dir, output_dir, artifact_name):
```
a-maurice, line 535: Can you add a comment describing what this function does?

a-maurice (approved):

PR #52, opened 2021-09-05
cleanup .gitignore

StephenHodgson:
- added `.vscode/`
- fixed spelling error for `*_unity_seperate/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ gcs_key_file.json
 *_build/
 *_unity/
 *_unity_seperate/
+*_unity_separate/
 
 # Windows ignore
 vcpkg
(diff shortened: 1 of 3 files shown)


StephenHodgson:
@googlebot I signed it!

.gitignore
```diff
@@ -20,7 +20,10 @@ gcs_key_file.json
 # Folders for cmake/test output
 *_build/
 *_unity/
-*_unity_separate/
+*_unity_seperate/
```
a-maurice, line 23: Thanks for sending this PR.  This is from some scripts, and also I believe it is spelt with an 'a'.  If you can undo this part though, should be good to add vscode

StephenHodgson: @a-maurice seems as though the scripts are busted. Not really sure where that'd be at but I guess I can do a search/replace. The problem is that there's probably some folks that have already cloned the repo and are suddenly going to expect the output folder in one location and it might create a duplicate. This at least keeps everything tidy.

a-maurice (approved):

PR #59, opened 2021-09-29
fix setup

cynthiajoan:
### Description
os setup fix
***

### Type of Change
***

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -14,9 +14,16 @@ jobs:
   build_ios:
   name: build-ios
     runs-on: macos-latest
-    unity_version: ${{ github.event.inputs.unity_version }}
-    unity_install_dir: /Applications/Unity_${{ github.event.inputs.unity_version }}
-    build_dir: ios_unity
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest]
+        include:
+        - os: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ios.yml
```diff
@@ -14,9 +14,16 @@ jobs:
   build_ios:
   name: build-ios
     runs-on: macos-latest
-    unity_version: ${{ github.event.inputs.unity_version }}
-    unity_install_dir: /Applications/Unity_${{ github.event.inputs.unity_version }}
-    build_dir: ios_unity
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest]
```
sunmou99, line 20: There is only one value here. Do you plan to expanded the value in the future?

.github/workflows/ios.yml
```diff
@@ -14,9 +14,16 @@ jobs:
   build_ios:
   name: build-ios
     runs-on: macos-latest
```
sunmou99, line 16: runs-on: ${{ matrix.os }}

.github/workflows/ios.yml
cynthiajoan: hmmm no, but not sure what's the best way to put extra attributes like unity version etc. The current setup has issue.

sunmou99 (approved):

PR #75, opened 2021-10-15
Add ability to specify Firebase CPP version

cynthiajoan:
### Description
Currently can setup to either use local SDK (pre downloaded) or specify a certain version (use versions set in the version cmake files)

Also update workflow to able to checkout a certain version or using latest
***
### Testing
Both locally and on workflow
***

### Type of Change
***


cmake/external/firebase_cpp_sdk.cmake
```diff
@@ -22,10 +22,15 @@ if("${FIREBASE_CPP_GIT_REPO}" STREQUAL "")
...
 
+if("${FIREBASE_CPP_SDK_VERSION}" STREQUAL "")
+  set(FIREBASE_CPP_SDK_VERSION ${FIREBASE_CPP_SDK_PRESET_VERSION})
+endif()
+
 ExternalProject_Add(
   firebase_cpp_sdk
 
   GIT_REPOSITORY ${FIREBASE_CPP_GIT_REPO}
+  GIT_TAG        v${FIREBASE_CPP_SDK_VERSION}
```
a-maurice, line 33: The 'v' should probably be part of the PRESET_VERSION, since that is described at the GitHub tag

a-maurice (dismissed):

cmake/external/firebase_cpp_sdk.cmake
cynthiajoan: DONE

PR #94, opened 2021-11-12
Integration Test workflow accept SDK artifact; and Packaging workflow can trigger Integration Test afterwards

sunmou99:
### Description
1. Integration Test workflow accept SDK artifact.
2. Packaging workflow can trigger Integration Test. Related scripts are copied from CPP repo and then modified for Unity repo.

not include:
fix build errors.
### Testing
[Packaging workflow](https://github.com/firebase/firebase-unity-sdk/actions/runs/1451514339) triggers [Integration test workflow](https://github.com/firebase/firebase-unity-sdk/actions/runs/1451530716).

***

### Type of Change
***


scripts/gha/github.py
```diff
@@ -0,0 +1,280 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A utility for GitHub REST API.
+
+This script handles GitHub Issue, Pull Request, Comment, Label and Artifact
```
cynthiajoan, line 17: NIT: the name is too general, maybe call github_rest_utils.py?

scripts/gha/trigger_workflow.py
```diff
@@ -0,0 +1,113 @@
...
+
+Usage:
+python trigger_workflow.py -w workflow_filename -t github_token [-b branch_name]
+       [-r git_repo_url] [-p <input1> <value1> -p <input2> <value2> ...]'
+       [-C curl_command]
+
+If -r is unspecified, uses the current repo.
+If -c is unspecified, uses the current HEAD.
+"""
+
```
cynthiajoan, line 30: Same comments for this file. Especially if we have more files that we already made in CPP, it's worth to checkout the cpp repo in the workflow and use the same script in these two repos.

scripts/gha/github.py
sunmou99: Thanks for the suggestion. 

Yes, we have a few similar tools in both CPP and Unity repo with subtle difference. 
I'm considering combining and moving these tools in a new repo (e.g. firebase-games-testing) in the future. But, right now the highest priority it's to make it work in Unity repo without breaking CPP CI.

Yes. I create [set_repo_url()](https://github.com/firebase/firebase-cpp-sdk/pull/746/files#diff-bd5d647c8c782da640a271074167ff54d565487e435358f3c5972c7ddea7e9c8R43) function in the script before I create this PR. 

I also don't see a big change of this files in the near future.

cynthiajoan (approved):

Issue #154, opened 2022-01-20
macOS Monterey < 12.3 = Python 2 deprecation popup | macOS Monterey >= 12.3 = Python error


julianxhokaxhiu:
Hi,

When using this package on Unity, while it imports it, Unity triggers a popup warning saying that the software is not ready for macOS future releases.

This behavior seems to be common across apps running Python scripts in general, like:
- https://github.com/microsoft/vscode/issues/134635
- https://grahamrpugh.com/2021/10/25/monterey-disable-python-2-deprecation-warnings.html

Can you please consider a solution towards this in order to make it future proof?

For anyone crossing this issue, the workaround is by running this command:
```sh
defaults write com.apple.python DisablePythonAlert True
```

Thank you in advance,
Julian

dustinkerstein:
And in the MacOS 15.3 Beta, Python2 has been removed from /usr/bin and this now causes errors for Firebase. For some reason Firebase can't use a symlinked Python3 in /usr/local/bin (which is accessible via PATH). I probably could symlink to /usr/bin, but that would require disabling/enabling SIP, and I think there probably needs to be a better solution. @julianxhokaxhiu, it might be worth updating the title of this issue as it's no longer just a deprecation warning; it now prevents the plugin from functioning correctly. The errors that I get are:

```
Unable to find command line tool python required for Firebase Android resource generation.
python is required to generate the Firebase Android resource file google-services.xml from Assets/GoogleService-Info.plist. Without Firebase Android resources, your app will fail to initialize.
python was distributed with each Firebase Unity SDK plugin, was it deleted?

System.ComponentModel.Win32Exception (0x80004005): ApplicationName='python', CommandLine='"/Users/dustin/GitHub/camera_unity/Assets/Firebase/Editor/generate_xml_from_google_services_json.py" -i "Assets/GoogleService-Info.plist" -l --plist', CurrentDirectory='/Users/dustin/GitHub/camera_unity', Native error= mono-io-layer-error (2)
  at System.Diagnostics.Process.StartWithCreateProcess (System.Diagnostics.ProcessStartInfo startInfo) [0x002dc] in <0463b2ef957545c0a51b42f372cd4fbb>:0 
  at System.Diagnostics.Process.Start () [0x0003a] in <0463b2ef957545c0a51b42f372cd4fbb>:0 
  at (wrapper remoting-invoke-with-check) System.Diagnostics.Process.Start()
[...]

PR #152, opened 2022-01-19
Build xcode testapp project on Windows and then build iOS testapp on macOS.

sunmou99:
[replace this line]: # (Describe your changes in detail.)
***

[replace this line]: # (Describe your testing in detail.)
***

### Type of Change
Place an `x` the applicable box:
- [ ] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [ ] Other, such as a build process or documentation change.
***

PR #176, opened 2022-02-15
Fix python issue on macOS 12.3 or later #152

sableangle:
Auto detect the OS family and OS Version and change the return value of `PYTHON_INTERPRETER` property (also change the member type of this member), because of the new version macOS(after 12.3) will no longer shipped the Python2.

***
> Describe how you've tested these changes.

Tested on macOS 12.3 beta2, macOS 12.2, with Unity Editor 2020.3.x, Unity 2021.2.x
Not tested on earlier macOS version and earlier Unity Editor version.

***

### Type of Change
***
Due to #154, the 'python' binary doesn't shipped with macOS after 12.3 (which is replace to 'python2.7' and 'python3').
So the Firebase.Editor script needs to change the binary file name if the macOS version is greater than 12.3. 
(Currently, just detect is the os version greater than 12 or not due to macOS 12 already shipped 'python3')

### DIFF
--- a/editor/app/src/PythonExecutor.cs
+++ b/editor/app/src/PythonExecutor.cs
@@ -124,7 +124,7 @@ namespace Firebase.Editor {
                 if (UnityEngine.SystemInfo.operatingSystemFamily == OperatingSystemFamily.MacOSX)
                 {
                     // Currently Unity API return Fixed format, so just remove that part, may have better solution
-                    var versionCodeString = UnityEngine.SystemInfo.operatingSystem.Replace("Mac OS X ", "").Split('.').First();
+                    var versionCodeString = UnityEngine.SystemInfo.operatingSystem.Replace("Mac OS X ", "").Split('.')[0];
                     // Default is 10, due to before Big Sur is 10.x
                     int versionCode = 10;
                     if (int.TryParse(versionCodeString, out versionCode))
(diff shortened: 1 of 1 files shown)


cynthiajoan (changes requested):
Thanks for the PR. Your branch seems behind which introduced a lot none related diffs. Can you please pull and merge first so we can focus on the change?

sableangle:
HI, I was developing after the 'main' branch on my fork repo.
I was original set the base branch to 'master' is due to the contributing guidelines suggest me do so.
But I found most of the PR is setting the base branch to 'main', so I also change the base branch to 'main' on this PR, and then all others commits the required to  pull and merge should gone.

I'm not sure is this action acceptable, but if it is then I think we can continue the pull requires proccesse.

sableangle (commented):
Auto detect the OS family and OS Version and change the return value of PYTHON_INTERPRETER

sableangle:
I also tested on macOS 11 with Unity2021.2.x
it works great!

sableangle:
Hi, macOS 12.3 is release today.
The old version SDK will soon unworkable on the Mac which update to the new version macOS.

Any suggestion on the PR? Any thing I can help to improve the python implementation, I will do the modify.

xysverma:
@cynthiajoan L Could it be taken forward. MacOS released today and i upgraded and i am blocked from using firebase due to this.

PR #191, opened 2022-03-09
make iOS build fat lib

cynthiajoan:
***

***

### Type of Change
***

### DIFF
--- a/build_ios.sh
+++ b/build_ios.sh
@@ -16,8 +16,83 @@
 #
 # Builds and runs the tests, meant to be used on a bash environment.
 
+usage(){
+    echo "Usage: $0 [options]
+ options:
+   -b, build path              default: ios_unity
+   -s, source path             default: .
+   -p, framework platform      default: ${SUPPORTED_PLATFORMS[@]}
+   -a, framework architecture  default: ${SUPPORTED_ARCHITECTURES[@]}
+ example: 
+   build_scripts/ios/build.sh -b ios_build -s . -a arm64"
+}
+
+set -e
+
+readonly SUPPORTED_PLATFORMS=(device) # simulator) only support device arch for now
+readonly SUPPORTED_ARCHITECTURES="arm64;armv7" #;x86_64;i386" only support device arch for now
+#readonly DEVICE_ARCHITECTURES="arm64;armv7" only support device arch for now
+#readonly SIMULATOR_ARCHITECTURES="arm64;x86_64;i386" only support device arch for now
+
+# build default value
+buildpath="ios_unity"
+sourcepath="."
+platforms=("${SUPPORTED_PLATFORMS[@]}")
+
 # Enable utf8 output
 export LANG=en_US.UTF-8
+
+# check options
+IFS=',' # split options on ',' characters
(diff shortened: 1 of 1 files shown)


build_ios.sh
```diff
@@ -16,9 +16,84 @@
...
+buildpath="ios_unity"
+sourcepath="."
+platforms=("${SUPPORTED_PLATFORMS[@]}")
+
 # Enable utf8 output
 export LANG=en_US.UTF-8
 
+# check options
+IFS=',' # split options on ',' characters
+while getopts ":b:s:a" opt; do
```
a-maurice, line 47: I believe the colon needs to be after the letter, as it is a way of saying the previous letter has options that go with it.  Also, missing p and h.  Also, can you add a c like I did for the android one, that just gets passed to the cmake configure step directly?

build_ios.sh
```diff
@@ -42,23 +117,16 @@ CMAKE_OPTIONS="${CMAKE_OPTIONS} -DUNITY_ROOT_DIR=${UNITY_ROOT_DIR}"
...
 # Make a directory to work in (if doesn't exist)
-mkdir -p "$DIR"
+mkdir -p "$buildpath"
 
-pushd "$DIR"
+pushd "$buildpath"
 
   # Configure cmake with option value
-  cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/unity_ios.cmake .. ${CMAKE_OPTIONS}
+  cmake -DCMAKE_TOOLCHAIN_FILE=${sourcepath}/cmake/unity_ios.cmake -DCMAKE_OSX_ARCHITECTURES=$SUPPORTED_ARCHITECTURES .. ${CMAKE_OPTIONS}
```
a-maurice, line 129: The .. means this is still assuming that the buildpath is one directory below the sourcepath.

build_ios.sh
cynthiajoan: Done

build_ios.sh
cynthiajoan: the buildpath is one directory below, move it to the front to be more clear.

build_ios.sh
```diff
@@ -16,9 +16,90 @@
...
+sourcepath="."
+platforms=("${SUPPORTED_PLATFORMS[@]}")
+cmake_extra=""
+
 # Enable utf8 output
 export LANG=en_US.UTF-8
 
+# check options
+IFS=',' # split options on ',' characters
+while getopts ":h:b:s:p:a:c" opt; do
```
a-maurice, line 49: This needs to be `hb:s:p:a:c:`

build_ios.sh
a-maurice: My concern is that buildpath is now a setting that can be passed in, so if the caller did something like "ios/x86", it would fail, because that is now 2 directories down. Maybe just a comment on the flag would be fine as well.

cynthiajoan: using ${sourcepath} to replace ".."

build_ios.sh
cynthiajoan: done

a-maurice (approved):

PR #220, opened 2022-04-08
Create firestore/scripts/cp_unity_testapp.py

dconeybe:
This script is a convenience script to copy the Firestore testapp to another directory and perform some helpful tweaks.


ehsannas (approved):
LGTM. There are a number of things here that depend on the exact path (directory/file) names, which will break if the testapp hierarchies change. I haven't seen that happen very often, and if it does we'll just need to update the script.

PR #231, opened 2022-04-15
Update the Dynamic Links testapp to handle Windows and Linux

a-maurice:
Dynamic Links uses Unity's Application.identifier under the hood to fill in some fields, but that results in an empty string on Windows and Linux. Since that is expected on desktop, change the tests to only check for that identifier on mobile and Apple platforms, which are expected to be present.
***

***

### Type of Change
***


a-maurice:
Checking it with https://github.com/firebase/firebase-unity-sdk/actions/runs/2174905474

a-maurice:
Re-running tests: https://github.com/firebase/firebase-unity-sdk/actions/runs/2185210720

cynthiajoan (approved):

PR #234, opened 2022-04-21
Add logic to use the special Android files for Messaging

a-maurice:
Messaging requires using a custom UnityPlayerActivity, which is meant to be included in the srcaar file. This adds the logic to use both the custom AndroidManifest.xml and classes.jar file. Note that this includes checking in the classes.jar file, which is a short term fix. Ideally, it would be generated as part of the build process, but more time is required to get that functional, as there are a handful of issues to resolve.
***

Local building and running of the testapp
***

### Type of Change
***


a-maurice:
Build tested with https://github.com/firebase/firebase-unity-sdk/actions/runs/2203893587

cynthiajoan (approved):

PR #262, opened 2022-04-29
Force on the CXX11_ABI flag

a-maurice:
The Linux Unity editor was crashing in the Firestore testapp. From what I can tell, the Unity editor was likely build with this option enabled, and mixing the library without it was the cause of the crash. So we will force the option to always be on for Linux builds. Note that standalone builds were not crashing, so it was likely just a problem with the Unity editor being built with that option.
***

***

### Type of Change
***

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,16 +130,13 @@ enable_language(C)
 enable_language(CXX)
 
 if(NOT MSVC AND NOT APPLE)
-  # Linux-specific option.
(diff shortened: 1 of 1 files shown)


a-maurice:
Desktop build: https://github.com/firebase/firebase-unity-sdk/actions/runs/2242859483

dconeybe:
Details about the `_GLIBCXX_USE_CXX11_ABI` preprocessor define are here: https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

PR #287, opened 2022-05-12
windows build python script

cynthiajoan:
### Description
Update the build python script to add support for windows, workflow not touched yet.
***

***

### Type of Change
***


scripts/build_scripts/build_zips.py
```diff
@@ -363,7 +363,7 @@ def get_windows_args():
       cmake args for windows platform.
   """
   result_args = []
-  result_args.append('-G \"Visual Studio 16 2019\"')
+  result_args.append('-G Visual Studio 16 2019')
```
DellaBitta, line 366: This documentation makes me feel like it should be 

```
'-G  \"Visual Studio 16 2019\"'
```

That is, it should continue to have the double quotes. No?

https://docs.microsoft.com/en-us/cpp/build/customize-cmake-settings?view=msvc-170

scripts/build_scripts/build_zips.py
```diff
@@ -562,8 +569,6 @@ def main(argv):
     cmake_setup_args.extend(get_ios_args(source_path))
   elif is_android_build():
     cmake_setup_args.extend(get_android_args())
```
DellaBitta, line 571: While Windows needs to be after the target path, is there any harm in these other platforms also being right after the target path?  That way the platform-specific if / else block is all in one place.

DellaBitta (changes requested):

.github/workflows/sdk_build.yml
```diff
@@ -148,6 +148,8 @@ jobs:
         if: startsWith(matrix.os, 'windows')
         shell: bash
         run: |
+          # swig_dir_slash_fix=${SWIG_DIR//\\//}
+          # python scripts/build_scripts/build_zips.py --platform=windows --unity_root="$UNITY_ROOT_DIR" --cmake_extras='-DSWIG_DIR=$swig_dir_slash_fix'
```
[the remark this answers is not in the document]
cynthiajoan: Yes currently don't want to change the release workflow, and the python scripts will be used in the new entire workflow chain.

scripts/build_scripts/build_zips.py
cynthiajoan: But the test outcome doesn't match what the doc says

DellaBitta (approved):

a-maurice (approved):

Issue #294, opened 2022-05-17
[Bug] Crashing on iOS 12.5.5


pf-mkenichi:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2020.3.32f1
  * Firebase Unity SDK version: 8.10.0, 8.10.1
  * Source you installed the SDK: FirebaseMessaging.unitypackage (.unitypackage or Unity Package Manager)
  * Problematic Firebase Component: Unknown (Auth, Database, etc.)
  * Other Firebase Components in use: not identified (Auth, Database, etc.)
  * Additional SDKs you are using: None (Facebook, AdMob, etc.)
  * Platform you are using the Unity editor on: Mac (Mac, Windows, or Linux)
  * Platform you are targeting: iOS (iOS, Android, and/or desktop)
  * Scripting Runtime: iL2CPP (Mono, and/or IL2CPP)
  * Pre-built SDK from the [website](https://firebase.google.com/download/unity) or open-source from this repo: https://firebase.google.com/download/unity

### [REQUIRED] Please describe the issue here:

Unity2020.3.32f1
Xcode version 13.3 
iPhone6Plus(iOS12.5.5)

With the above configuration, import FirebaseMessaging.unitypackage and run
When using Firebase 8.10 or later version, the app crashes on startup. 
Occurs with just adding firebase to an empty project.

Firebase 8.9.0 No problem
Firebase 8.10.0 Crash
Firebase 8.10.1 crash

Not reproduced on iPhoneX (iOS 15.4)

#### Steps to reproduce:

Have you been able to reproduce this issue with just the Firebase Unity quickstarts (this GitHub project)?
I haven't tried it.

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

xcode log
-----------
2022-05-17 13:00:41.939976+0900 firebase-sample[1231:505132] FCM: Loading UIApplication FIRFCM category
(lldb) 
-----------

xcode screenshot iPhone6Plus(iOS12.5.5)

<img width="1680" alt="xcode-screen-shot" src="https://user-images.githubusercontent.com/29224244/168727162-ebc1261d-6266-4dd7-bd7e-782eec8de24d.png">

acgourley:
We're seeing this too. We didn't QA on this device class before release and now we have justifiable upset users 😞

acgourley:
[...]

PR #297, opened 2022-05-18
Set the iOS minimum version flag in the toolchain

a-maurice:
Set CMAKE_OSX_DEPLOYMENT_TARGET, which in turn sets the iOS minimum version flag. Note though that it seems a bit buggy when building fat libraries, as it gets confused on which minimum version flag to use. This will work on clean builds, which will be enough for a release, but for a long term fix we will likely need to change the build scripts to build device and simulator separately, and then merge them together afterwards.
***

Building locally, and running on an iOS simulator with iPhoneOS 12.4, and running on a device with iPhoneOS 13.3
***

### Type of Change
***

https://github.com/firebase/firebase-unity-sdk/issues/294

https://github.com/firebase/quickstart-unity/issues/1275

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -160,10 +160,10 @@ Release Notes
     - General: Minimum supported editor version is now Unity 2018.
     - General (Editor, macOS): Add support for Apple Silicon chips.
     - General (iOS): Firebase Unity on iOS is now built using Xcode 13.3.1.
+    - General (iOS): Fixed crash when running on iPhoneOS 12 and older.
     - Analytics: Removed deprecated event names and parameters.
     - Crashlytics (Android): Fixed a bug with missing symbols when enabling
       minification via proguard.
-    - Messaging (iOS): Fixed crash when running on iPhoneOS 12 and older.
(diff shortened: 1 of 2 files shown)


docs/readme.md
```diff
@@ -163,6 +163,7 @@ Release Notes
     - Analytics: Removed deprecated event names and parameters.
     - Crashlytics (Android): Fixed a bug with missing symbols when enabling
       minification via proguard.
+    - Messaging (iOS): Fixed crash when running on iPhoneOS 12 and older.
```
dconeybe, line 166: nit: This applies to Firestore as well (and possibly other products).

cynthiajoan (approved):

docs/readme.md
a-maurice: I was waffling on that.  The issue we were seeing with the RC missing stream symbols wasn't present in 8.10.1.  The issue in 8.10.1 seems to do with how messaging hooks into the delegates, but it is possible this did affect other products in some way as well in 8.10.1, so I'll move it to general.

cynthiajoan (approved):

PR #301, opened 2022-05-19
Fix messaging duplicate symbol bug

a-maurice:
In the transition to GHA, Messaging lost the extra dependency on iid it needs, or else there is a problem when also using Functions with duplicated symbols.
***

***

### Type of Change
***

https://github.com/firebase/quickstart-unity/issues/1274

### DIFF
--- a/cmake/android_dependencies.cmake
+++ b/cmake/android_dependencies.cmake
@@ -60,9 +60,11 @@ set(FIREBASE_INSTALLATIONS_ANDROID_DEPS
     "com.google.firebase:firebase-analytics:21.0.0"
 )
 
+# iid is needed by messaging to avoid a conflict with functions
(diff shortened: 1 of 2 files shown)


cmake/android_dependencies.cmake
```diff
@@ -73,4 +75,4 @@ set(FIREBASE_REMOTE_CONFIG_ANDROID_DEPS
 set(FIREBASE_STORAGE_ANDROID_DEPS
     "com.google.firebase:firebase-storage:20.0.1"
     "com.google.firebase:firebase-analytics:21.0.0"
-)
```
cynthiajoan, line 76: extra line

cynthiajoan: nvm... you are actually fixing the extra line.

cynthiajoan (approved):

PR #305, opened 2022-05-19
Add playmode (in editor mode) test in CI

sunmou99:
### Description
`Playmode` added as a new platform.

***
### Testing
See the comment below
***

### Type of Change
***

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -18,8 +18,8 @@ on:
         default: 'macos-latest'
         required: true
       platforms:
-        description: 'CSV of Android,iOS,Windows,macOS,Linux'
-        default: 'Android,iOS,Windows,macOS,Linux'
+        description: 'CSV of Android,iOS,Windows,macOS,Linux,Playmode'
+        default: 'Android,iOS,Windows,macOS,Linux,Playmode'
         required: true
       apis:
         description: 'CSV of apis to build and test'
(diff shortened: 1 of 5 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -393,6 +393,139 @@ jobs:
...
+      LC_ALL: en_US.UTF-8
+      LANG: en_US.UTF-8
+      U3D_PASSWORD: ""
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Unity installer (U3D)
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          gem install u3d
```
cynthiajoan, line 417: do they fix the u3d bugs? We have to ping to an older version to get installation correct

sunmou99: I think so: https://github.com/DragonBox/u3d/issues/421

cynthiajoan: Hmm we tested 1.3.1 and it still fails, that's the reason we currently pinged to 1.2.3
https://github.com/firebase/firebase-unity-sdk/pull/300

sunmou99: But the tests was running successfully: https://github.com/firebase/firebase-unity-sdk/runs/6516557068?check_suite_focus=true#step:6:1
Probably because playmode only requires the basic unity installation.

cynthiajoan (approved):

PR #313, opened 2022-05-24
Add GetHttpsCallableFromURL to Functions

a-maurice:
Adds GetHttpsCallableFromURL to the Functions SDK. This was recently added to both the Android and iOS SDKs.
***

Running locally
***

### Type of Change
***

Will require https://github.com/firebase/firebase-cpp-sdk/pull/964, and updating the firebase_cpp default

### DIFF
--- a/functions/src/FirebaseFunctions.cs
+++ b/functions/src/FirebaseFunctions.cs
@@ -222,6 +222,26 @@ namespace Firebase.Functions {
     }
 
     /// <summary>
+    ///   Creates a
+    ///   <see cref="HttpsCallableReference" />
+    ///   given a URL.
+    /// </summary>
+    public HttpsCallableReference GetHttpsCallableFromURL(string url) {
+      ThrowIfNull();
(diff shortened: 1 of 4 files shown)


functions/src/FirebaseFunctions.cs
```diff
@@ -221,6 +221,26 @@ public sealed class FirebaseFunctions {
       return new HttpsCallableReference(this, functionsInternal.GetHttpsCallable(name));
     }
 
+    /// <summary>
+    ///   Creates a
```
cynthiajoan, line 225: hmmm this seems an extra line, can the reference fit in the same line?

cynthiajoan (dismissed):

functions/src/FirebaseFunctions.cs
a-maurice: I copied it from above, but you are right, there isn't a need for it to be so broken up. Fixed the above one as well.

PR #320, opened 2022-05-31
cp_unity_testapp.py: add --hardlink command-line flag

dconeybe:


firestore/scripts/cp_unity_testapp.py
```diff
@@ -142,6 +152,7 @@ def __init__(self) -> None:
     self.google_service_info_plist_file: Optional[pathlib.Path] = None
     self.android_package_name: Optional[str] = None
     self.apple_developer_team_id: Optional[str] = None
+    self.hardlink_cs_files: Optional[bool] = None
```
ehsannas, line 155: I'm confused about whether the intention is for this to be optional or not.

dconeybe: It's only optional because it gets initialized to None, then later on it gets set to `FLAG_HARDLINK_CS_FILES.value`. I suppose I can just initialize it to `FLAG_HARDLINK_CS_FILES.value` and make it non-optional. I'll do that since it's less confusing.

ehsannas (approved):

PR #325, opened 2022-06-01
Firestore: cp_dlls.py added

dconeybe:
Add a script useful during development to copy the build artifacts into an existing Unity project.

I've personally found this script very useful, so I'm publishing it into the official repository in hopes that others will find it useful as well.

The script only support macOS and Linux right now. I expect to add Windows, Android, and iOS support in the future.


cp_dlls.py
```diff
@@ -0,0 +1,244 @@
...
+
+    # Do not delete old CPP app library files if we aren't even copying them
+    # due to the copy_pattern not matching the file name.
+    if self.copy_pattern is not None \
+        and not self.copy_pattern.fullmatch(src_cpp_app_library_file_name):
+      return
+
+    # TODO(dconeybe) Update this logic to also support Mac M1, which probably
+    # uses a different directory than "x86_64".
+    x86_64_dir = self.dest_dir / "Assets" / "Firebase" / "Plugins" / "x86_64"
```
ehsannas, line 183: TIL about `/`! I think I used `path.join()` in my past life

ehsannas (approved):

PR #348, opened 2022-06-08
Handle passing null to cpp instance manager

a-maurice:
The internal GetInstance classes are sometimes expected to return null, for example Storage does if passing an invalid bucket.  Instead of checking for null in each case, it seems better to let the cpp_instance_manager handle when null is passed in, especially instead of storing those nulls, as it currently does, assuming assertions are off.
***

Building and running locally
***

### Type of Change
***


jonsimantov (approved):
Do we check the result anywhere?

a-maurice:
> Do we check the result anywhere?

Only place I saw was in the test file (that is currently not used).

PR #394, opened 2022-07-06
Add link xml file for Crashlytics

a-maurice:
When using il2cpp, the compiler more aggressively strips unused symbols, which incorrectly removes Crashlytics unless the user calls into it, which isn't meant to be required.
***

Building locally to validate the file is placed in the zip file correctly.
***

### Type of Change
***

https://github.com/firebase/quickstart-unity/issues/1290


crashlytics/link.xml
```diff
@@ -0,0 +1,5 @@
+<linker>
+<assembly fullname="Firebase.Crashlytics" ignoreIfMissing="1">
+    <namespace fullname="Firebase.Crashlytics" preserve="all"/>
+</assembly>
+</linker>
```
cynthiajoan, line 5: need extra line?

cynthiajoan (approved):

crashlytics/link.xml
a-maurice: Fixed

PR #423, opened 2022-07-27
update one product build package logic

cynthiajoan:
***

***

### Type of Change
***


unity_packer/debug_single_export_json/analytics.json
```diff
@@ -0,0 +1,397 @@
+{
```
a-maurice, line 1: Are all of these just the usual export, but with just App and the product they care about?  It feels bad having to maintain an extra set of files for this.

.github/workflows/package.yml
```diff
@@ -49,6 +50,22 @@ jobs:
     strategy:
       fail-fast: false
     steps:
+      - name: Check input
```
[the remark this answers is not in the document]
cynthiajoan: done

scripts/build_scripts/build_package.py
```diff
@@ -141,6 +178,22 @@ def main(argv):
   if os.path.exists(output_folder):
     shutil.rmtree(output_folder)
 
+  if FLAGS.apis:
+    # This should only happen for debug purpose, to create packages for certain products
+    # Codes below should not run
```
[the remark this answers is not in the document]
cynthiajoan: if the apis are passed in, it should skip the rest of the packaging steps going below. I can also do an else if that feels more exclusive.

unity_packer/debug_single_export_json/analytics.json
cynthiajoan: Yes the export jsons are like what you described. I used a script to generate those so should not be too much maintain work. I also upload that script if that's better.

a-maurice: I think it is fine for now. In the future we can look into making it use the main one, but we do rarely touch these files in general.

scripts/build_scripts/build_package.py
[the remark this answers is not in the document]
a-maurice: Maybe just rephrase it to:
If told to only build a subset, package just those products and exit early.

.github/workflows/package.yml
```diff
@@ -49,6 +50,24 @@ jobs:
...
+        # This check the apis input. It detect whether the package is only asking for subset of
+        # of the products, like auth,firestore only etc. If the input apis are less than the
+        # default_apis env attribute, we determine this package only package subset and for 
+        # check up purpose only.
```
a-maurice, line 57: Instead of "we determine..."  maybe "we package only the provided apis, for test purposes only."

scripts/build_scripts/build_package.py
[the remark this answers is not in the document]
cynthiajoan: done

.github/workflows/package.yml
cynthiajoan: done

a-maurice (approved):

PR #436, opened 2022-08-04
Uninstall Android platforms that fail with Unity

a-maurice:
The Firestore Android testapp was failing to build, because it requires minification. The root cause is that the GitHub runners have the android-33 platform installed, and since it is the latest one, Unity tries to build with it, but some of the libraries (android.car.jar as an example) was build with Java 11.  This requires a newer JDK and Gradle than Unity uses, which causes the build to fail.  As a workaround, we can uninstall android-33 (the runners have 27 through 33 installed).
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/2792702607
***

### Type of Change
***

### DIFF
--- a/scripts/gha/build_testapps.py
+++ b/scripts/gha/build_testapps.py
@@ -527,6 +527,16 @@ def patch_android_env(unity_version):
       logging.info("Uninstall Android build tool 31.0.0")
     except Exception as e:
       logging.info(str(e))
+
+  try:
+    # The platform android-33 includes libraries that were built with Java 11, and require a newer version of gradle
+    # than Unity comes with. Note this only happens when using minification.
+    # If this continues to be a problem, this logic might need to be smarter, to remove all versions newer than 32,
+    # but currently the GitHub runners have 33 as their max.
+    logging.info("Uninstall Android platform android-33")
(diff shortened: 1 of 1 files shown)


scripts/gha/build_testapps.py
```diff
@@ -527,6 +527,16 @@ def patch_android_env(unity_version):
...
+
+  try:
+    # The platform android-33 includes libraries that were built with Java 11, and require a newer version of gradle
+    # than Unity comes with. Note this only happens when using minification.
+    # If this continues to be a problem, this logic might need to be smarter, to remove all versions newer than 32,
+    # but currently the GitHub runners have 33 as their max.
+    logging.info("Uninstall Android platform android-33")
+    _run([os.environ["ANDROID_HOME"]+"/tools/bin/sdkmanager", "--uninstall", "platforms;android-33"], check=False)
+  except Exception as e:
+    logging.info(str(e))
```
dconeybe, line 539: drive-by comment: consider using `logging.exception()` instead of `logging.info` so that the whole stack trace is logged.

sunmou99 (approved):
LGTM

PR #465, opened 2022-09-20
tvOS build capability added to build_zips

DellaBitta:
Breaking ground on tvOS support.
This PR updates the cmake build system and the build_zips.py script to generate fat libs for tvOS simulators and devices.

***
### Testing

- Local builds: `python scripts/build_scripts/build_zips.py --platform=tvos`
- Confirmed multi arch support by executing `lipo -info` on the resulting static libraries.
- Confirmed the existance of symbols by running `nm -S` on the resulting libraries.

***

### Type of Change
***


scripts/build_scripts/build_zips.py
```diff
@@ -17,7 +17,7 @@
 
 
 Example usage:
-  python build_zips.py --platform=macos --targets=auth --targets=firestore
+  python build_zips.py --platform=macos --apis=auth --targets=firestore
```
a-maurice, line 20: Technically it takes either targets or apis (I'm fine with this change, just providing info)

scripts/build_scripts/build_zips.py
```diff
@@ -167,14 +186,19 @@ def get_targets_args(targets):
       camke args included targets.
   """
   result_args = []
+  
+  support_targets = SUPPORT_TARGETS
+  if is_tvos_build() and not targets:
+    targets = TVOS_SUPPORT_TARGETS
```
a-maurice, line 192: For tvOS do you also want to set support_targets to TVOS_SUPPORT_TARGETS, to catch if someone tries to run: --platform=tvos --apis=messaging
?

scripts/build_scripts/build_zips.py
```diff
@@ -625,6 +785,8 @@ def main(argv):
     logging.info("Build macos with multiple architectures %s",
                  ",".join(g_target_architectures))
     make_macos_multi_arch_build(cmake_setup_args)
+  elif is_tvos_build:
```
a-maurice, line 788: missing ()?

scripts/build_scripts/build_zips.py
```diff
@@ -167,14 +186,25 @@ def get_targets_args(targets):
...
   """
   result_args = []
+  
+  support_targets = SUPPORT_TARGETS
+  if is_tvos_build():
+    support_targets = TVOS_SUPPORT_TARGETS
+    # Explicitly disable Dynamic Links on tvOS. This is required because we're
+    # truncating support_targets to a subset, and the "target in support_targets"
+    # loop below won't append this property definition for us.
+    result_args.append("-DFIREBASE_INCLUDE_DYNAMIC_LINKS=OFF")
```
a-maurice, line 196: I think it would make more sense to enforce this in the CMake logic, like how C++ handled it: https://github.com/firebase/firebase-cpp-sdk/blob/main/CMakeLists.txt#L120

a-maurice (approved):

PR #470, opened 2022-09-22
Deprecate `PhoneAuthProvider.MaxTimeoutMs` (Fixed #54)

### LINKED ISSUE #54: Make PhoneAuthProvider.MaxTimeoutMs settable since 3 seconds is too short
When calling .verifyPhoneNumber, I can set a autoVerifyTimeoutMS parameter, which, according to the docs, is clamped to MaxTimeoutMs property of PhoneAuthProvider. This MaxTimeoutMs property is 3 seconds, which is - at least on my two devices - just too short. On both my Android devices it takes about 5 seconds after codeSent until the actual sms is received. 

Unfortunately, MaxtimeoutMs is only gettable, not settable. I would like to extend this value to 10 seconds so that I can make use of autoVerification. 

Right now my users would have to copy and paste the 6-digit-code into an input field, which could - at least in my eyes - easily "fixed" by allowing to set this autoVerifyTimeout.

Or am I missing something?

FirebaseUnity 8.1.0
Unity 2020.3.16f1

chkuang-g:
In response to a Unity issue #54

This is a read-only valued used to set the upper bound for `autoVerifyTimeOutMs`.  However, the value was set too low. Android SDK has increase the range to 30 to 120 seconds, while iOS SDK does not have a way to override timeout.

This is a change to deprecate this public property so that the underlying SDK can govern the range for timeout.

Depends on C++ PR: https://github.com/firebase/firebase-cpp-sdk/pull/1084

***

N/A

***

### Type of Change
***


auth/src/PhoneAuthProvider.cs
```diff
@@ -57,13 +57,6 @@ namespace Firebase.Auth {
 ///       send an SMS. Credential is automatically created and passed to
 ///       the app via @ref VerificationCompleted.
 public sealed class PhoneAuthProvider : global::System.IDisposable {
-  /// Maximum value of `autoVerifyTimeOutMs` in @ref VerifyPhoneNumber.
-  /// @ref VerifyPhoneNumber will automatically clamp values to this amount.
-  public static uint MaxTimeoutMs {
-    get {
-      return PhoneAuthProviderInternal.kMaxTimeoutMs;
-    }
-  }
```
DellaBitta, line 66: As mentioned in the C++ SDK, it's better to deprecate this field instead of removing it as this is a breaking change.

DellaBitta (changes requested):

cynthiajoan (approved):

auth/src/PhoneAuthProvider.cs
chkuang-g: Alright. Added them back. Used both `System.Obsolete` and `@deprecated` for both compiler warning and ref doc

DellaBitta (approved):

PR #478, opened 2022-10-03
Unity Editor support for tvOS pods and xcode projects

DellaBitta:
Updates to the Unity build process to handle tvOS pod dependencies and xcode project file configuration.

***

- Local builds on macOS for macOS, iOS and tvOS.
- [Integration Test CI](https://github.com/firebase/firebase-unity-sdk/actions/runs/3183486870)
- [Unity SDK Build CI](https://github.com/firebase/firebase-unity-sdk/actions/runs/3183497310)

***

### Type of Change
***

### DIFF
--- a/app/platform/Unity/PlatformInformation.cs
+++ b/app/platform/Unity/PlatformInformation.cs
@@ -31,7 +31,8 @@ internal static class PlatformInformation {
   // Is the current platform iOS?
   internal static bool IsIOS {
     get {
-      return UnityEngine.Application.platform == UnityEngine.RuntimePlatform.IPhonePlayer;
+      return UnityEngine.Application.platform == UnityEngine.RuntimePlatform.IPhonePlayer ||
+         UnityEngine.Application.platform == UnityEngine.RuntimePlatform.tvOS;
     }
   }
 
--- a/app/platform/Unity/PlatformInformation.cs
+++ b/app/platform/Unity/PlatformInformation.cs
@@ -28,7 +28,7 @@ internal static class PlatformInformation {
     }
   }
 
-  // Is the current platform iOS?
+  // Is the current platform iOS or tvOS?
   internal static bool IsIOS {
     get {
       return UnityEngine.Application.platform == UnityEngine.RuntimePlatform.IPhonePlayer ||
--- a/editor/app/src/XcodeProjectPatcher.cs
+++ b/editor/app/src/XcodeProjectPatcher.cs
@@ -64,20 +64,23 @@ internal class XcodeProjectPatcher : AssetPostprocessor {
     private static bool Enabled {
         get {
             return (EditorUserBuildSettings.activeBuildTarget ==
-                    BuildTarget.iOS) && Google.IOSResolver.Enabled;
+                    BuildTarget.iOS ||
+                    EditorUserBuildSettings.activeBuildTarget ==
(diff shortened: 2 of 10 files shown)


firestore/testapp/Assets/Firebase/Editor/Builder.cs
```diff
@@ -43,6 +43,23 @@ public class Builder {
     Build(options);
   }
 
+  public static void BuildTvos() {
```
cynthiajoan, line 46: nit: it looks a bit wired with the capital choice, same to the previous Ios, but I don't know what's the best way to deal with it :(

cynthiajoan (approved):

firestore/testapp/Assets/Firebase/Editor/Builder.cs
DellaBitta: Yeah, I'm not a huge fan either. I looked it up and I _think_ this is actually in line with [Google's C# style guide](https://google.github.io/styleguide/csharp-style.html). Methods are Pascal Case and only acronyms of 2 letters or less are fully capitalized.

DellaBitta: It still makes my eyes hurt, though. :)

editor/app/src/XcodeProjectPatcher.cs
```diff
@@ -96,15 +99,24 @@ internal class XcodeProjectPatcher : AssetPostprocessor {
...
+    private static string GetApplicationId() {
+        switch(EditorUserBuildSettings.activeBuildTarget) {
+            case BuildTarget.iOS:
+                return UnityCompat.GetApplicationId(BuildTarget.iOS);
+                break;
+            case BuildTarget.tvOS:
+                return UnityCompat.GetApplicationId(BuildTarget.tvOS);
+                break;
+            default:
+                throw new Exception("unsupported iOS+ version");
```
a-maurice, line 112: This is a noticeable change of behavior.  Are there any chances that GetApplicationId gets called when the BuildTarget isn't one of those two?  It feels like it would be safer to return tvOS is set to that, otherwise return iOS.

editor/crashlytics/src/iOSPostBuild.cs
```diff
@@ -69,18 +69,19 @@ public class iOSPostBuild {
     [PostProcessBuild(100)]
     public static void OnPostprocessBuild(BuildTarget buildTarget, string buildPath) {
       // BuiltTarget.iOS is not defined in Unity 4, so we just use strings here
```
a-maurice, line 71: Remove old comment about Unity 4

app/platform/Unity/PlatformInformation.cs
```diff
@@ -31,7 +31,8 @@ internal static class PlatformInformation {
   // Is the current platform iOS?
```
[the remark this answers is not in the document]
DellaBitta: Done.

editor/app/src/XcodeProjectPatcher.cs
DellaBitta: Ok. Done. I also changed the method name to `GetIosPlusApplicationId`

editor/crashlytics/src/iOSPostBuild.cs
DellaBitta: Done.

a-maurice (approved):

PR #480, opened 2022-10-06
Add a tvOS build workflow

DellaBitta:
First pass at a workflow to build the Unity SDK for tvOS platforms.

***

Manaully executed the python scripts which would be invoked by the workflow.

***

### Type of Change
***


.github/workflows/build_tvos.yml
```diff
@@ -0,0 +1,154 @@
+# Workflow to handle building the Unity SDK on tvOS
+name: Build tvOS (SubWorkflow)
+
+on:
+  workflow_dispatch:
```
cynthiajoan, line 5: I think we don't need this for subworkflow, we can keep it for test purpose, but once it's hooked with the build_starter.yml, we need to remove this block.

.github/workflows/build_tvos.yml
```diff
@@ -0,0 +1,154 @@
...
+        type: string
+      unity_platform_name:
+        description: 'The platform name Unity should install with'
+        default: ''
+        required: true
+        type: string
+
+jobs:
+  build_tvos:
+    name: build-tvos-unity${{ inputs.unity_version }}-CPP${{ inputs.firebase_cpp_sdk_version }}
```
cynthiajoan, line 61: echo to the previous comment, this would not take effect if the workflow start with dispatch.

.github/workflows/build_tvos.yml
```diff
@@ -0,0 +1,154 @@
...
+        timeout-minutes: 90
+        shell: bash
+        run: |
+          # TODO add handling cmake_extras
+          python scripts/build_scripts/build_zips.py --gha --platform=tvos --unity_root=$UNITY_ROOT_DIR --apis=${{ inputs.apis }}
+
+      - name: Check zip file
+        shell: bash
+        run: |
+          if [ -f ios_unity/*.zip ]; then
```
cynthiajoan, line 143: is the build_zip python script update to create tvos_unity folder? Not sure i've seen the related PR yet.

.github/workflows/build_tvos.yml
DellaBitta: Yes, that's for me to iterate on.

.github/workflows/build_tvos.yml
DellaBitta: Yep!

.github/workflows/build_tvos.yml
DellaBitta: Confused by this comment. Both the dispatch and the trigger have these inputs, no?

.github/workflows/build_tvos.yml
cynthiajoan: Great, if that's the case the check needs to update to if [ -f tvos_unity/*.zip ]

.github/workflows/build_tvos.yml
cynthiajoan: My experience is the workflow_dispatch use another parameter name for the input, needs to be events.inputs something like that. workflow_call is using "inputs" directly. examples can be found build_starter.yml, - id: set_outputs step.

cynthiajoan (approved):
LGTM, just prefer to have the check build success or not step's path corrected from ios_unity to tvos_uniy

DellaBitta:
Oh yes, I missed the `ios_unity` -> `tvos_unity`  in the bash if statement. Will fix before merge. Thanks!

a-maurice (approved):
Approved assuming the ios -> tvos fix that was already called out.

.github/workflows/build_tvos.yml
DellaBitta: I think it should be ok, based on this reference:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

PR #484, opened 2022-10-12
Add workflow that builds tvOS

DellaBitta:
Fixed up workflow to build Unity for tvOS targets.
Workflow is triggered by manual triggers still - and not by the `Firebase Unity SDK build` which kicks-off builds for all of the platforms. This is simply so we don't release the tvOS libraries. Later, when the tvOS release is ready, we'll add this `Build tvOS (SubWorkflow)` to the main build list.

***

- [Build tvOS CI](https://github.com/firebase/firebase-unity-sdk/actions/runs/3234991120)
- Inspected resulting artifact to ensure it contains unity libraries for all of the corresponding products, and that they're in the correct paths.
- Executed `lipo -info` on the resulting static libraries to ensure that they contain the intended architectures of `x86_64` for simulator targets and `arm64` for AppleTV device targets.

***

### Type of Change
***

### DIFF
--- a/.github/workflows/build_tvos.yml
+++ b/.github/workflows/build_tvos.yml
@@ -130,9 +130,25 @@ jobs:
         run: |
           echo "UNITY_ROOT_DIR=$( python scripts/gha/print_matrix_configuration.py -u ${{ inputs.unity_version }} -k unity_path )" >> $GITHUB_ENV
 
-      - name: Install prerequisites
+      - name: Pod repo list (pre)
+        shell: bash
+        run: pod repo list
+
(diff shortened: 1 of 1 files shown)


DellaBitta:
Fixed up workflow to build Unity for tvOS targets.

Workflow is triggered by manual triggers still - and not by the `Firebase Unity SDK build` which kicks-off builds for all of the platforms. This is simply so we don't release the tvOS libraries. Later, when the tvOS release is ready, we'll add this `Build tvOS (SubWorkflow)` to the main build list.

PR #504, opened 2022-10-25
Fix Android integration testapp build

a-maurice:
The Android SDK has a couple of different paths the sdkmanager could be, so attempt those when searching for it.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/3317150614
***

### Type of Change
***


sunmou99 (approved):
LGTM

PR #507, opened 2022-10-27
Firestore instructions and convinience script

wu-hui:


.gitignore
```diff
@@ -30,3 +30,49 @@ vcpkg
 
 # Visual Studio Code
 .vscode/
+
+# Firestore testapp
+firestore/testapp/**/*.dll
```
dconeybe, line 35: Drive-by comment: Consider just creating a .gitignore in the `firestore` subdirectory, rather than putting Firestore-specific entries in the top-level .gitignore.

firestore/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
+# Overview
+
+This describes how Firestore Unity SDK works, and how to develop and test 
+the SDK, targeting desktop/Android/iOS.
+
+# Prerequisites
+
+Building the Unity SDK requires building the underlying C++ SDK. Refer to
+[][this doc] for what the prerequisites are.
```
dconeybe, line 9: nit: missing link

firestore/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+
+On top of above, you also need Unity installed (obviously). If you use an
+apple silicon machine as host, be sure to install the right version of
```
dconeybe, line 12: nit: Explain what you mean by "right version", and what happens if you install the "wrong version".

firestore/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+
+Building Firestore into Unity Packages is a very involved process, mixed with
+multiple build tools working together. Therefore, we will rely on Python scripts
+to automate the process. The scripts live under `$REPO_ROOT/scripts/build_scripts`.
+
+```zsh
+# all scripts are run from the repo root.
+
+# Building for Mac. The build tools will try to find Unity automatically
+python scripts/build_scripts/build_zips.py -platform=macos -targets=auth -targets=firestore -use_boringssl
```
dconeybe, line 25: Do you have to use boringssl? It generally gives faster build times to use a precompiled openssl, such as the one installed via homebrew.

firestore/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+cp -r ./scripts/gha/integration_testing/automated_testapp/ftl_testapp_files firestore/testapp/Assets/Firebase/Sample/
+```
+
+To run the test app, open `firestore/testapp` from Unity Editor, and load the Unity packages we built above.
+Then open up `firestore/testapp/Assets/Firebase/Sample/Firestore/MainSceneAutomated.unity`, you should be
+able to run this scene which in turn runs all integration tests for Firestore.
+
+# Running Firestore Android TestApp
+
+You *probably* need to use `IL2CPP` as scripting backend instead of `Mono` for Android. To do this,
```
dconeybe, line 66: Please explain why you `probably need to use IL2CPP`. For example, why would Mono not work, or be a bad choice?

firestore/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+you can go to `Edit->Project Setting->Player->Android->Scripting Backend` and select `IL2CPP`.
+
+You also need to turn on `minification` under on the same setting page, by turning on `R8` under `publish
```
dconeybe, line 69: Explain that if you fail to turn on minification you run into the "Too many field references... max is 65536" issue. An alternative is to configure multidex (https://developer.android.com/studio/build/multidex,

dconeybe (commented):
I just left a few drive-by comments

wu-hui (commented):
Thanks for driving-by!

PR #520, opened 2022-11-03
Remove python3.11 when doing the mac build

a-maurice:
The mac runners are having issues with python3.11.  This is fixed downstream, but to unblock the build remove the python3.11 install.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/3388914873
(The failure is because auth wasn't turned on, but it got past the python problem)
***

### Type of Change
***


dconeybe:
Note that if this is on account of a failing build of Firestore then this was fixed by https://github.com/firebase/firebase-ios-sdk/pull/10440, which isn't yet included in a release.

AlmostMatt:
@dconeybe Yea this is a temporary workaround until the iOS fix is included in a release

AlmostMatt (approved):

PR #529, opened 2022-11-08
Fix the GHA tests for Android

a-maurice:
Because of some issue with the test runner, it keeps failing to build Crashlytics and Messaging on Android.  The root cause for this seems to be that the Editor dlls, and some dlls from EDM4U, fail to load correctly due to iOS support libraries being missing. While they should not be necessary, until we can track down the root cause this workaround will let the tests build again.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/3415146709
***

### Type of Change
***

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -103,7 +103,8 @@ jobs:
           pip install -r scripts/gha/requirements.txt
       
       - id: unity_setup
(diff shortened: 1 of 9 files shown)


DellaBitta (approved):

sunmou99:
Could you share the reason why the retry logic has been removed from the script?

PR #549, opened 2022-11-28
Implement Crashlytics Unity ODFs

mrober:
Implement Crashlytics Unity ODFs for iOS and Android.

Design doc: [go/crashlytics-unity-fatals](http://go/crashlytics-unity-fatals)
API review: [go/crashlytics-unity-fatals-api](http://go/crashlytics-unity-fatals-api)

### DIFF
--- a/crashlytics/src/AndroidImpl.cs
+++ b/crashlytics/src/AndroidImpl.cs
@@ -145,22 +145,22 @@ namespace Firebase.Crashlytics
     public override void LogExceptionAsFatal(Exception exception) {
       var loggedException = LoggedException.FromException(exception);
       Dictionary<string, string>[] parsedStackTrace = loggedException.ParsedStackTrace;
-      
+
       if (parsedStackTrace.Length == 0) {
         // if for some reason we don't get stack trace from exception, we add current stack trace in
         var currentStackTrace = System.Environment.StackTrace;
         LoggedException loggedExceptionWithCurrentStackTrace = new LoggedException(loggedException.Name, loggedException.Message, currentStackTrace);
         parsedStackTrace = loggedExceptionWithCurrentStackTrace.ParsedStackTrace;
 
-        if (parsedStackTrace.Length > 2) {
-          // remove RecordCustomException and System.Environment.StackTrace frame for fault blame on crashlytics sdk
-          var slicedParsedStackTrace = parsedStackTrace.Skip(2).Take(parsedStackTrace.Length - 2).ToArray();
+        if (parsedStackTrace.Length > 3) {
+          // remove AndroidImpl frames for fault blame on crashlytics sdk
+          var slicedParsedStackTrace = parsedStackTrace.Skip(3).Take(parsedStackTrace.Length - 3).ToArray();
           parsedStackTrace = slicedParsedStackTrace;
         }
       }
-  
+
       StackFrames frames = new StackFrames();
-      foreach (Dictionary<string, string> frame in loggedException.ParsedStackTrace) {
+      foreach (Dictionary<string, string> frame in parsedStackTrace) {
         frames.Add(new FirebaseCrashlyticsFrame {
           library = frame["class"],
           symbol = frame["method"],
--- a/crashlytics/src/AndroidImpl.cs
+++ b/crashlytics/src/AndroidImpl.cs
@@ -141,6 +141,24 @@ namespace Firebase.Crashlytics
         }, "LogException");
     }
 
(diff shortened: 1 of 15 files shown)


crashlytics/src/cpp/android/crashlytics_android.cc
```diff
@@ -62,19 +62,25 @@ METHOD_LOOKUP_DEFINITION(firebase_crashlytics,
...
+
 #define CRASHLYTICS_CORE_FIELDS(X)                                           \
   X(DataCollectionArbiter, "dataCollectionArbiter",                          \
   "Lcom/google/firebase/crashlytics/internal/common/DataCollectionArbiter;", \
   util::kFieldTypeInstance)
 
 // clang-format on
 METHOD_LOOKUP_DECLARATION(crashlytics_core,
-                          METHOD_LOOKUP_NONE, CRASHLYTICS_CORE_FIELDS)
+                          CRASHLYTICS_CORE_METHODS,
```
cynthiajoan, line 77: what the reason of making this change?

crashlytics/src/AndroidImpl.cs
```diff
@@ -141,6 +141,24 @@ internal class AndroidImpl : Impl {
...
+      StackFrames frames = new StackFrames();
+      foreach (Dictionary<string, string> frame in loggedException.ParsedStackTrace) {
+        frames.Add(new FirebaseCrashlyticsFrame {
+          library = frame["class"],
+          symbol = frame["method"],
+          fileName = frame["file"],
+          lineNumber = frame["line"],
+        });
+      }
+      // TODO(mrober): Do something for empty stacktrace. Get current stacktrace?
```
cynthiajoan, line 155: Do we still need this TODO? If so can we clarify the comment more? By reading I'm not quite sure what is supposed to happen.

cynthiajoan (approved):
LGTM in general with some minor questions.

crashlytics/src/cpp/android/crashlytics_android.cc
mrober: The logFatalException method is an internal api of CrashlyticsCore, not a public api of Crashlytics. So we added it here so we can call it via jni.

crashlytics/src/AndroidImpl.cs
themiswang: @mrober tried without stack trace and there is no blame for log only on demand fatals: https://screenshot.googleplex.com/tAZaCTidi3NaWxo

crashlytics/src/AndroidImpl.cs
```diff
@@ -141,6 +142,39 @@ internal class AndroidImpl : Impl {
         }, "LogException");
     }
 
+    public override void LogExceptionAsFatal(Exception exception) {
+      var loggedException = LoggedException.FromException(exception);
+      Dictionary<string, string>[] parsedStackTrace = loggedException.ParsedStackTrace;
+      
+      if (parsedStackTrace.Length == 0) {
```
themiswang, line 149: Hey @cynthiajoan, we have addressed the TODO previously we had. If the stack trace from exception is empty, we will parse the current stack trace in with two frames less to prevent false blame on  crashlytics sdk. Do you mind check this logic and lgtm again, thank you!

crashlytics/src/AndroidImpl.cs
```diff
@@ -141,6 +142,39 @@ internal class AndroidImpl : Impl {
...
+        // if for some reason we don't get stack trace from exception, we add current stack trace in
+        var currentStackTrace = System.Environment.StackTrace;
+        LoggedException loggedExceptionWithCurrentStackTrace = new LoggedException(loggedException.Name, loggedException.Message, currentStackTrace);
+        parsedStackTrace = loggedExceptionWithCurrentStackTrace.ParsedStackTrace;
+
+        if (parsedStackTrace.Length > 2) {
+          // remove RecordCustomException and System.Environment.StackTrace frame for fault blame on crashlytics sdk
```
themiswang, line 156: @mrober I directly copy this block from iOSUmpl.cs and then realized Android does not have `RecordCustomException` this function. Can you confirm if removing 1 stack frame is good enough for android?

mrober: I tuned it a bit in 92a567, before: https://screenshot.googleplex.com/4JnexnXpYZ6dbzE.png after: https://screenshot.googleplex.com/4qc8Fy48ZUfQKpG.png

cynthiajoan (approved):

themiswang (approved):

PR #598, opened 2023-01-13
[crashlytics ondemand fatals]fix the fatal logic

themiswang:
### Description
Fix iOS fatal for on demand
Add current stack trace if there is empty stack trace in exception

***
### Testing
Tested exception with stack trace:
https://screenshot.googleplex.com/43Grf4RVMDd5Fzn

Tested exception without stack trace:
https://screenshot.googleplex.com/BNWeCg59nHA46Lt
***

### DIFF
--- a/crashlytics/src/IOSImpl.cs
+++ b/crashlytics/src/IOSImpl.cs
@@ -20,6 +20,7 @@ namespace Firebase.Crashlytics
   using System.Runtime.InteropServices;
   using System.Diagnostics;
   using System.Collections.Generic;
+  using System.Linq;
   using UnityEngine;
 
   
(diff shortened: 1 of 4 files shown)


crashlytics/src/IOSImpl.cs
```diff
@@ -116,6 +117,19 @@ public override void LogException(Exception exception)
     private void RecordCustomException(LoggedException loggedException, bool isOnDemand) {
       Dictionary<string, string>[] parsedStackTrace = loggedException.ParsedStackTrace;
 
+      if (isOnDemand && parsedStackTrace.Length == 0) {
+        // if for some reason we don't get stack trace from exception, we add current stack trace in
+        var currentStackTrace = System.Environment.StackTrace;
+        LoggedException loggedExceptionWithCurrentStackTrace = new LoggedException(loggedException.Name, loggedException.Message, currentStackTrace);
+        parsedStackTrace = loggedExceptionWithCurrentStackTrace.ParsedStackTrace;
+
+        if (parsedStackTrace.Length > 2) {
```
themiswang, line 126: If we don't add this logic, the stack trace will look like this as it still blame on crashlytics frame:
Screen Shot 2023-01-13 at 12 27 19 AM

Remove these two frame for logging purpose then here is the result:
Screen Shot 2023-01-13 at 12 26 43 AM

PR #626, opened 2023-02-10
Package the Messaging java file and aar separately

a-maurice:
Instead of baking the Messaging java file and C++ aar into the srcaar file, keep them separate. This is to support Unity 2023, which has modified the behavior of UnityPlayerActivity, specifically the mUnityPlayer member variable.  Note that additional work is still needed for the new UnityPlayerGameActivity logic.

Depends on https://github.com/firebase/firebase-cpp-sdk/pull/1205
***

Tested locally with 2019.4.40f1, 2021.3.13f1, and 2023.1.0b3.
https://github.com/firebase/firebase-unity-sdk/actions/runs/4180060512
***

### Type of Change
***


cmake/android_dependencies.cmake
```diff
@@ -65,6 +65,7 @@ set(FIREBASE_MESSAGING_ANDROID_DEPS
     "com.google.firebase:firebase-messaging:23.1.1"
     "com.google.firebase:firebase-analytics:21.2.0"
     "com.google.firebase:firebase-iid:21.1.0"
+    "com.google.flatbuffers:flatbuffers-java:1.12.0"
```
cynthiajoan, line 68: Are we going to update the version if flatbuffer has a new release?

cmake/unity_pack.cmake
```diff
@@ -308,19 +309,26 @@ endfunction()
...
   if("${UNITY_PACK_PACK_PATH}" STREQUAL "")
     set(UNITY_PACK_PACK_PATH ${UNITY_PACK_DEFAULT_CS_PATH})
   endif()
 
-  # meta file
-  install(
-    FILES ${file}
-    DESTINATION ${UNITY_PACK_PACK_PATH}
-  )
+  if ("${UNITY_PACK_RENAME_TO}" STREQUAL "")
```
cynthiajoan, line 320: Can we add some comments about this section?

a-maurice: Added some more comments.

cmake/android_dependencies.cmake
a-maurice: We'll keep this in sync with whatever version the C++ repo is using.

cynthiajoan (approved):

PR #635, opened 2023-02-24
Fix user-agent register timing in Unity SDK

chkuang-g:
Make sure the user-agents are registered before C++ App is created.

Depends on a new utility function introduced in https://github.com/firebase/firebase-cpp-sdk/pull/1217

***

E2E test using Auth iOS/Android testapp
***

### Type of Change
***

### DIFF
--- a/app/src/swig/app.i
+++ b/app/src/swig/app.i
@@ -51,7 +51,7 @@
 
 #if defined(__ANDROID__)
 #include "app/src/google_play_services/availability_android.h"
-#include "app/src/util_android.h"
+#include "app/src/app_android.h"
 #endif // defined(__ANDROID__)
 
 namespace firebase {
@@ -173,7 +173,7 @@ static void RegisterLibrariesHelper(
 #if FIREBASE_PLATFORM_ANDROID
     JNIEnv* jni_env;
     jobject activity_local_ref = UnityGetActivity(&jni_env);
-    firebase::util::CallAfterEnsureMethodsCached(
+    firebase::CallAfterEnsureMethodsCached(
             jni_env, activity_local_ref, [&libraries, &jni_env](){
         for (std::map<std::string, std::string>::const_iterator it =
                 libraries.begin(); it != libraries.end(); ++it) {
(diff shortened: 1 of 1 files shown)


app/src/swig/app.i
```diff
@@ -167,6 +168,49 @@ static void LogHeartbeatForDesktop(App* app) {
 #endif  // FIREBASE_PLATFORM_DESKTOP
 }
 
+static void RegisterLibraryHelper(const char* library, const char* version) {
```
AlmostMatt, line 171: Can we delete RegisterLibraryInternal and RegisterLibraryHelper?
It looks like there is no current usage of RegisterLibraryInternal.

Alternatively, if RegisterLibraryInternal is still useful, can it be changed to just create a map with a single entry and then call RegisterLibrariesInternal?

That way we would not need to have these two very similar helpers.

app/src/swig/app.i
```diff
@@ -832,6 +878,27 @@ static firebase::AppOptions* AppOptionsLoadFromJsonConfig(const char* config) {
     Firebase.Platform.FirebaseHandler.Create(
         Firebase.Platform.FirebaseAppUtils.Instance);
     FirebaseApp newProxy;
+
+    // If this is the first App, register library information.
+    if (!userAgentRegistered) {
+        userAgentRegistered = true;
+        var userAgentMap = new StringStringMap();
```
AlmostMatt, line 885: since 'new' is used here, I think the map should be deleted after the call to RegisterLibrariesInternal

AlmostMatt (commented):
This looks pretty good to me.

app/src/swig/app.i
chkuang-g: This is C# code block so userAgentMap would be simply GCed.

app/src/swig/app.i
chkuang-g: Good call. Deleted.

a-maurice (approved):

jonsimantov (approved):

PR #641, opened 2023-02-28
Update .gitignore and patch packaging script

chkuang-g:
Update `.gitignore` to ignore downloaded ndk and EDM4U
Also update packaging script to check if EDM4U is downloaded to `*_unity` folder.

***

N/A
***

### Type of Change
***

### DIFF
--- a/scripts/build_scripts/build_package.py
+++ b/scripts/build_scripts/build_package.py
@@ -116,7 +116,8 @@ def find_pack_script():
   built_folder = None
   resolver_root_folder = "unity-jar-resolver"
   for folder in os.listdir("."):
-    if folder.endswith(built_folder_ext):
+    if (folder.endswith(built_folder_ext) and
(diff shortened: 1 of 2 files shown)


scripts/build_scripts/build_package.py
```diff
@@ -116,7 +116,8 @@ def find_pack_script():
   built_folder = None
   resolver_root_folder = "unity-jar-resolver"
   for folder in os.listdir("."):
-    if folder.endswith(built_folder_ext):
+    if ( folder.endswith(built_folder_ext) and
```
cynthiajoan, line 119: nit, don't need extra space

cynthiajoan (approved):

scripts/build_scripts/build_package.py
chkuang-g: Done.
Also removed one debug message

Issue #557, opened 2022-12-01
[Bug] Build this repo failed on windows


Thaina:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2022.1.16f1
  * Firebase Unity SDK version: latest from repo
  * Source you installed the SDK: git repo
  * Problematic Firebase Component: source code
  * Other Firebase Components in use: -
  * Additional SDKs you are using: -
  * Platform you are using the Unity editor on: Windows
  * Platform you are targeting: Android
  * Scripting Runtime: -
  * Pre-built SDK from the [website](https://firebase.google.com/download/unity) or open-source from this repo: This repo

#### Steps to reproduce:

I just clone this repo and run 

`python scripts/build_scripts/build_zips.py --platform=android`

And it just got error below

#### Relevant Code:

```
cmd /C "C:\Python311\python.exe c:\Users\Brothers_Model\.vscode\extensions\ms-python.python-2022.18.2\pythonFiles\lib\python\debugpy\adapter/../..\debugpy\launcher 65148 -- C:\Users\Brothers_Model\Documents\git\firebase-unity-sdk/scripts/build_scripts/build_zips.py --platform=android "
I1201 15:47:22.586176 20768 build_zips.py:307] Use ANDROID_NDK_HOME(C:\Program Files\Unity\Hub\Editor\2022.1.16f1\Editor\Data\PlaybackEngines\AndroidPlayer\NDK) cmake toolchain(C:\Program Files\Unity\Hub\Editor\2022.1.16f1\Editor\Data\PlaybackEngines\AndroidPlayer\NDK\build\cmake\android.toolchain.cmake)
I1201 15:47:22.587175 20768 build_zips.py:779] cmake_setup_args is: cmake C:\Users\Brothers_Model\Documents\git\firebase-unity-sdk -DFIREBASE_INCLUDE_UNITY=ON -DFIREBASE_UNITY_BUILD_TESTS=ON -DFIREBASE_CPP_BUILD_STUB_TESTS=ON -DCMAKE_TOOLCHAIN_FILE=C:\Program Files\Unity\Hub\Editor\2022.1.16f1\Editor\Data\PlaybackEngines\AndroidPlayer\NDK\build\cmake\android.toolchain.cmake -DFIREBASE_ANDROID_BUILD=true -DCMAKE_BUILD_TYPE=release -DANDROID_STL=c++_shared
[...]

PR #646, opened 2023-03-07
fix for build windows

Thaina:
- Explicitly disable boringssl to use openssl 
- Add netstandard reference to make build success
- Fix the call to `firebase::App::RegisterLibrary` which only have 2 parameter for windows platform

More details in https://github.com/firebase/firebase-unity-sdk/issues/557

***

Run `python ./scripts/build_scripts/build_zips.py --use_boringssl=false --platform=windows` and got zip and dll files result properly without error

***

### Type of Change
- [ x ] Bug fix. Add the issue # below if applicable.
- [ x ] Other, such as a build process or documentation change.
***

ps This is cmakesettings.json I need to explicitly configure in my machine

```json
{
  "configurations": [
    {
      "name": "x64-Debug",
      "generator": "Ninja",
      "configurationType": "Debug",
      "inheritEnvironments": [ "msvc_x64_x64" ],
      "buildRoot": "${projectDir}\\out\\windows_unity\\${name}",
      "installRoot": "${projectDir}\\out\\install\\${name}",
      "cmakeCommandArgs": "",
      "buildCommandArgs": "",
      "ctestCommandArgs": "",
      "variables": [
        {
          "name": "UNITY_ROOT_DIR",
          "value": "C:/Program Files/Unity/Hub/Editor/2022.2.9f1/",
          "type": "STRING"
        },
        {
          "name": "MONO_EXE",
          "value": "C:/Program Files/Unity/Hub/Editor/2022.2.9f1/Editor/Data/MonoBleedingEdge/bin/mono.exe",
          "type": "STRING"
        },
        {
          "name": "MONO_CSHARP_BUILD_EXE",
          "value": "C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools/MSBuild/Current/Bin/MSBuild.exe",
          "type": "STRING"
        },
        {
          "name": "UNITY_MONO_EXE",
          "value": "C:/Program Files/Unity/Hub/Editor/2022.2.9f1/Editor/Data/MonoBleedingEdge/bin/mono.exe",
          "type": "STRING"
        },
        {
          "name": "UNITY_CSHARP_BUILD_EXE",
          "value": "C:/Program Files/Unity/Hub/Editor/2022.2.9f1/Editor/Data/MonoBleedingEdge/bin/xbuild.bat",
          "type": "STRING"
        },
        {
          "name": "ZLIB_INCLUDE_DIR",
          "value": "C:/Program Files (x86)/GnuWin32/include/",
          "type": "STRING"
        },
        {
          "name": "ZLIB_LIBRARY",
          "value": "C:/Program Files (x86)/GnuWin32/lib/zlib.lib",
          "type": "STRING"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/cmake/unity_mono.cmake
+++ b/cmake/unity_mono.cmake
@@ -275,6 +275,7 @@ macro(mono_add_internal name output_type)
   set(VAR_DEFINES "TRACE")
   set(VAR_OUTPUT_TYPE ${output_type})
 
+  string(APPEND VAR_REFERENCE "    <Reference Include=\"netstandard\" />\n")
   string(APPEND VAR_REFERENCE "    <Reference Include=\"System\" />\n")
   string(APPEND VAR_REFERENCE "    <Reference Include=\"System.Core\" />\n")
 
--- a/scripts/build_scripts/build_zips.py
+++ b/scripts/build_scripts/build_zips.py
@@ -771,6 +771,8 @@ def main(argv):
 
   if FLAGS.use_boringssl:
     cmake_setup_args.append("-DFIREBASE_USE_BORINGSSL=ON")
+  else:
+    cmake_setup_args.append("-DFIREBASE_USE_BORINGSSL=OFF")
 
   if is_ios_build():
     cmake_setup_args.extend(get_ios_args(source_path))


app/src/swig/app.i
```diff
@@ -191,8 +191,7 @@ static void RegisterLibrariesHelper(
         const std::string& library = it->first;
         const std::string& version = it->second;
         firebase::App::RegisterLibrary(library.c_str(),
-                                       version.c_str(),
-                                       nullptr);
+                                       version.c_str());
```
a-maurice, line 194: This was a recent change to the Firebase C++ SDK, so you might just be out of date on that.  https://github.com/firebase/firebase-cpp-sdk/pull/1217/files#diff-fa379ca24a0012efe3cc69026f1be97dcba3f5883f45c26cb5314f4b2c3190ee

a-maurice (changes requested):
Thanks for the contribution.  In general it seems fine, just one minor thing

app/src/swig/app.i
Thaina: I see, we need to update subrepo too

But then, are there anyway we could also update subrepo along with the main repo?

a-maurice (approved):

PR #649, opened 2023-03-09
Disable Session Id test on Android

a-maurice:
The session id test on Android can randomly fail due to the test infrastructure having potentially old versions of Play Services.  Disable the test on Android, and just rely on the combination of the C++ tests, and the other platforms to guarantee that it is calling into the C++ layer correctly.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/4379166466
***

### Type of Change
***


jonsimantov (approved):

jonsimantov:
Another option would be to still call GetSessionId on Android, but just to ensure it doesn't crash.

a-maurice:
> Another option would be to still call GetSessionId on Android, but just to ensure it doesn't crash (and not care about whether it succeeded).

Yeah, I switched over to that, feels a bit better. Also logs that is the case, which can be useful

PR #656, opened 2023-03-16
Generate the Messaging activity, and support GameActivity

a-maurice:
Add an editor script that generates the special Messaging activity java file, instead of including it directly. This is to better support the new GameActivity option, by changing the base class the Messaging activity overrides.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/4504885342
***

### Type of Change
***


messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+    // Only run this logic when building for Android.
+    if (EditorUserBuildSettings.activeBuildTarget != BuildTarget.Android) {
+      return;
+    }
+
+    // Check if the file has already been generated.
+    string[] oldAssetGuids = AssetDatabase.FindAssets("l:" + GeneratedFileTag);
+    if (oldAssetGuids != null && oldAssetGuids.Length > 0) {
+      if (oldAssetGuids.Length != 1) {
+        Debug.LogWarning("FirebaseMessagingActivityEditor found multiple generated files with the label: " +
```
chkuang-g, line 138: I think the user does not have enough information to act upon this warning message.

I assume this happens when the user move `FirebaseMessagingActivityEditor.java` to a different folder for management purpose.  If there are multiple copies, there could be a risk that the build can fail due to duplicated classes. Here are some potential way I can think of to make this easier for the user:

1. Keep the warning. But list down the path of duplicated files and suggest them to remove the duplication.
2. Change this to `info` level`. 
3. We just remove all the files with this tag if there are more than 2 of them, and regenerate it again. This would force the user to place the file only at the designated location.

messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+        return;
+      }
+      string oldAssetPath = AssetDatabase.GUIDToAssetPath(oldAssetGuids[0]);
+      Object oldAsset = AssetDatabase.LoadMainAssetAtPath(oldAssetPath);
+      // If the generated file has been tagged to be preserved, don't change it.
+      string[] labelList = AssetDatabase.GetLabels(oldAsset);
+      if (labelList.Contains(PreserveTag)) {
+        return;
+      }
+      // Delete the old asset.
```
chkuang-g, line 149: I think we should also check if the content of the file is the same to what the code will inject.

- We should reduce the frequency to trigger asset database change as much as possible since that can spin up a whole lot of other process reacting to asset database changes.  
- If the new file is using a different GUID, this is also pretty annoying for version controlling.
- This is also a good chance to check if the user changed the `GameActivity` flag.

messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+      AssetDatabase.DeleteAsset(oldAssetPath);
```
chkuang-g, line 150: I think we should notify the user about the deletion of the file with a `LogInfo` here and let them know the use of `PreserveTag` in the log in case they are modifying the file themselves.

Here I assume we only delete and regenerate when the file content are different.

messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+"}}"
+  };
+  private readonly string BaseActivityClass = "UnityPlayerActivity";
+  private readonly string BaseGameActivityClass = "UnityPlayerGameActivity";
+
+  private readonly string GeneratedFileTag = "FirebaseMessagingActivityGenerated";
+  // If this tag is present on the generated file, it will not be replaced.
+  private readonly string PreserveTag = "FirebasePreserve";
+
+  private readonly string OutputPath = Path.Combine("Plugins", "Android");
```
chkuang-g, line 124: I have a feeling you need to put Android related files under a folder name ended with `.androidlib`, ex. `Assets/Plugins/Android/FirebaseMessaging.androidlib`.

https://docs.unity3d.com/Manual/android-library-project-import.html

chkuang-g (changes requested):

messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+// Handles the generation of the MessagingUnityPlayerActivity java file.
+// Note this regenerates the file every time an Android build occurs,
+// but local changes can be preserved by using the PreserveTag below.
+// This is needed because the source code needs to be present to work across
+// different Unity versions, due to changes in mUnityPlayer.
+// It also adjusts the base class of the file based on if GameActivity is being
+// used (a new feature in Unity 2023).
+public class FirebaseMessagingActivityGenerator : IPreprocessBuildWithReport {
+  private readonly string[] ActivityClassContents = new string[]{
+"/*",
```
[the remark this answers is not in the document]
a-maurice: I tried that originally, the problem I had was consistently locating the template file.  When it is in the Assets folder, that is simple, but it can potentially be in the Packages folder, and I couldn't find a clear way to locate it in both cases.

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: I believe the only way this would happen would be if the user adds the label to additional files.  If they move it, it should still only find the one file with the label, so that should be fine.  I guess they could potentially copy the file themselves, which might lead to problems.  I can expand on the error message.

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: I think that is only if you are using a prebuilt library, or need an Android Manifest file.  For pure java files, it just needs to be present and set to Android: https://docs.unity3d.com/Manual/android-java-and-kotlin-plugins-create.html

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: That's a good point.  Done.

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: Done

messaging/activity/FirebaseMessagingActivityGenerator.cs
chkuang-g: The instruction mentioned that you need to enable `Android` in platform, which can be done in `.meta` metadata file.

Have you try to build the project and see if this `FirebaseMessagingActivityEditor.java` is included in the generated Gradle project? You can also use export project (a build option), which should save you time to find the generated Gradle project under `Library` folder.

messaging/activity/FirebaseMessagingActivityGenerator.cs
chkuang-g: Yeah. Like you said, there can be two cases:

1. The user accidentally label unrelated assets with this tag, which can be pretty common.  But this is harmless. There is a simple way to rule out this case by using `string.EndWith("MessagingUnityPlayerActivity.java")`
2. If the user copy `MessagingUnityPlayerActivity.java` to a different folder. This case _may_ cause build fail.

I think the minimum we can do is to provide the solution in the log.  Ex. in the end of your log, adds `"Please consider removing duplicated {OutputFilename}, or remove the tag "{PreserveTag}" from unintended assets."`

WDYT?

messaging/activity/FirebaseMessagingActivityGenerator.cs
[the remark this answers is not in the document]
a-maurice: Yeah, I imagine refactoring it will take a bit of time.  I added a TODO, cause I would much rather use a template file in the longer term though.

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: Added a bit more to the log message.

messaging/activity/FirebaseMessagingActivityGenerator.cs
a-maurice: Yeah, from all my tests Unity detects it is a java file, and automatically sets it up to just be included in the Android build.  This worked for both building directly in Unity, and exporting to Gradle.

messaging/activity/FirebaseMessagingActivityGenerator.cs
```diff
@@ -0,0 +1,183 @@
...
+    // If using the new GameActivity logic, we want to generate with that base class.
+    if (PlayerSettings.Android.applicationEntry.HasFlag(AndroidApplicationEntry.GameActivity)) {
+      baseClass = BaseGameActivityClass;
+    }
+#endif
+    string fileContents = System.String.Format(System.String.Join("\n", ActivityClassContents), baseClass);
+
+    // Check if the file has already been generated.
+    string[] oldAssetGuids = AssetDatabase.FindAssets("l:" + GeneratedFileTag);
+    if (oldAssetGuids != null && oldAssetGuids.Length > 0) {
```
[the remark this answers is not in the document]
a-maurice: Haha, of course.  Added a check.

chkuang-g (approved):

PR #667, opened 2023-03-30
Uninstall platforms;android-33-ext5 when building testapps

AlmostMatt:
### Description
This is effectively the same as #604, but for another extension of android 33.
Without this change build_testapps fails the minify with proguard step for build-2020-windows-latest-Android-NA: example: https://github.com/firebase/firebase-unity-sdk/actions/runs/4559389559/jobs/8043295718

### Testing
This is the same fix as for a previous version of the same issue.
***

### Type of Change
***


AlmostMatt:
Note that since this same issue has occurred twice, a better long-term fix is preferred, but this should fix the builds for now.

PR #669, opened 2023-04-05
Add initial App Check logic

a-maurice:
Add the initial build logic and swig file for App Check. Note that it isn't being added to the GitHub workflows yet, to prevent it from being included in the release until it is ready.
***

Building App Check locally.
https://github.com/firebase/firebase-unity-sdk/actions/runs/4624523264
***

### Type of Change
***


cynthiajoan:
do you need to add empty.cc under app_check/ ?

cynthiajoan (approved):

a-maurice:
> do you need to add empty.cc under app_check/ ?

The logic uses the empty.cc file at the top level, firebase_unity_sdk.  Really, I think we could probably get rid of all the random extra ones we have lying around, but that is lower priority.

PR #672, opened 2023-04-11
Setup initial App Check testapp

a-maurice:
This is copied from the Storage testapp.  The changes are:
 - Renaming the folder Assets/Firebase/Sample/Storage -> AppCheck
 - Clearing out most of the contents of UIHandler.cs and UIHandlerAutomated.cs
 - Updating the Application identifier in ProjectSettings.asset
 - Updating the readme
***

Building it locally
***

### Type of Change
***


AlmostMatt (approved):
Looks ok. I wonder if readme might need some additional instructions for setting up AppCheck in the Firebase console, but that can be added after the testapp has more logic.

PR #674, opened 2023-04-13
Add implementation for FirebaseAppCheck class

a-maurice:
Adds logic for GetAppCheckTokenAsync, to call into the C++ SDK, and also logic for the C++ SDK to call up to the Factory and Provider set in Unity.
Adds TokenChanged event, that gets callbacks when the App Check token is changed.
***

Testing localling
***

### Type of Change
***

### DIFF
--- a/app_check/src/AppCheckToken.cs
+++ b/app_check/src/AppCheckToken.cs
@@ -29,7 +29,7 @@ public struct AppCheckToken {
   public string Token { get; set; }
 
   /// The time at which the token will expire.
-  public System.DateTime ExpireTime { get; set; }
+  public DateTime ExpireTime { get; set; }
 
   internal static readonly DateTime s_unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
--- a/app_check/src/AppCheckToken.cs
+++ b/app_check/src/AppCheckToken.cs
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+using System;
+
 namespace Firebase.AppCheck {
 
 /// @brief Token used by the Firebase App Check service.
@@ -28,6 +30,24 @@ public struct AppCheckToken {
 
   /// The time at which the token will expire.
   public System.DateTime ExpireTime { get; set; }
+
+  internal static readonly DateTime s_unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+  // Construct an AppCheckToken given the swig generated class for one.
+  internal static AppCheckToken FromAppCheckTokenInternal(AppCheckTokenInternal tokenInternal) {
+    return new AppCheckToken {
+      Token = tokenInternal.token,
+      ExpireTime = s_unixEpoch.AddMilliseconds((double)tokenInternal.expire_time_millis)
+    };
+  }
+
+  // Get the expire time in milliseconds since Epoch, which is used by C++.
+  internal long ExpireTimeMs {
+    get {
+      TimeSpan ts = ExpireTime - s_unixEpoch;
+      return (long)ts.TotalMilliseconds;
+    }
+  }
 }
 
 }
--- a/app_check/src/FirebaseAppCheck.cs
+++ b/app_check/src/FirebaseAppCheck.cs
@@ -90,7 +90,9 @@ public sealed class FirebaseAppCheck {
     if (appCheckFactory == factory) return;
 
     appCheckFactory = factory;
-    // Clear out the Providers that were previously made
+    // Clear out the Providers that were previously made. When future calls to
+    // GetToken fails to find a provider in the map, it will use the new factory
+    // to create a new provider.
     providerMap.Clear();
 
     // Register the callback for C++ SDK to use that will reach this factory.
--- a/app_check/src/FirebaseAppCheck.cs
+++ b/app_check/src/FirebaseAppCheck.cs
@@ -32,10 +32,13 @@ public sealed class FirebaseAppCheck {
   private static IAppCheckProviderFactory appCheckFactory;
   private static Dictionary<string, IAppCheckProvider> providerMap =
     new Dictionary<string, IAppCheckProvider>();
+
+  // Function for C++ to call when it needs to fetch a Token.
   private static AppCheckUtil.GetTokenFromCSharpDelegate getTokenDelegate =
     new AppCheckUtil.GetTokenFromCSharpDelegate(GetTokenFromCSharpMethod);
-
-  private static readonly DateTime s_unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+  // Function for C++ to call when the Token changes.
+  private static AppCheckUtil.TokenChangedDelegate tokenChangedDelegate =
+    new AppCheckUtil.TokenChangedDelegate(TokenChangedMethod);
 
   // Make the constructor private, since users aren't meant to make it.
   private FirebaseAppCheck(AppCheckInternal internalObject) {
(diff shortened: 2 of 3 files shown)


app_check/src/FirebaseAppCheck.cs
```diff
@@ -95,11 +113,93 @@ public sealed class FirebaseAppCheck {
...
+  }
+
+  [MonoPInvokeCallback(typeof(AppCheckUtil.GetTokenFromCSharpDelegate))]
+  private static void GetTokenFromCSharpMethod(string appName, int key) {
+    if (appCheckFactory == null) {
+      AppCheckUtil.FinishGetTokenCallback(key, "", 0,
+        (int)AppCheckError.InvalidConfiguration,
+        "Missing IAppCheckProviderFactory.");
+    }
+    FirebaseApp app = FirebaseApp.GetInstance(appName);
```
AlmostMatt, line 164: Just a comment:

On iOS and Android there were some edge cases where the native SDK would try to create a provider and get a token before the C++ app had fully initialized. I wonder if something similar will happen before C# app is fully initialized. Perhaps that would result in a second FirebaseApp being created or some other issue.

That said, this is fine for now, but if such an issue comes up when we get to testing AppCheck, something might need to change.

app_check/src/swig/app_check.i
```diff
@@ -20,14 +20,187 @@
...
+    }
+  }
+
+ private:
+  App* app_;
+};
+
+// C++ implementation of the AppCheckProviderFactory that can
+// forward to the C# implementation.
+class SwigAppCheckProviderFactory : public AppCheckProviderFactory {
```
AlmostMatt, line 108: The way that these C++ and C# factories and providers interact was not intuitive to me.
I think I am starting to understand.

For iOS and Android the pattern that I followed was native provider and factory wrapping C++ provider and factory one to one. So native factory's CreateProvider calls C++ factory's CreateProvider and then creates a native provider to wrap the newly created C++ provider.

If I understand correctly, there is a single C++ SwigAppCheckProviderFactory which has no reference to the C# factories. It creates one SwigAppCheckProvider for each App, but these C++ providers have no reference to C# providers. Instead GetToken for these providers calls into a single static g_get_token_from_csharp method which looks up C# providers by name.

The way that the C# providers are used is that there is a map from app name to C# provider, and when the GetToken is called it finds the corresponding C# provider in the map.
The way that the C# factories work are used is that when the static C# factory changes, the provider map is cleared. When it next tries to lookup providers in that map it will use the new C# factory to create a new C# provider.

I think this pattern is probably fine. I'm assuming that the benefit of this pattern is a smaller number of classes and methods that call back and forth through the C++ and C# layers.

It will probably be good to add some comments to clarify the responsibility of each of these classes to help future readers of the code.

For example
SwigAppCheckProviderFactory
// This C++ factory is independent of C# factories. The current C# factory is stored in the variable appCheckFactory.

SwigAppCheckProvider
// This C++ provider is independent of C# providers. Logic to select the corresponding C# provider happens later in GetTokenFromCSharpMethod.

GetTokenFromCSharpMethod
// Given an app, create a C# provider if one does not already exist, and call GetToken on the C# provider.

app_check/src/swig/app_check.i
```diff
@@ -20,14 +20,187 @@
...
+    return provider;
+  }
+ private:
+  std::map<App*, AppCheckProvider*> provider_map_;
+};
+
+static SwigAppCheckProviderFactory g_swig_factory;
+
+// Called from C# to register the C# to call to get a token.
+void SetGetTokenCallback(GetTokenFromCSharp get_token_callback) {
```
AlmostMatt, line 139: Why do we need a method to set g_get_token_from_csharp?
Can it not always be equal to getTokenDelegate?
In fact, can we remove one of g_get_token_from_csharp or getTokenDelegate?
(It would be fine to still set the native factory to g_swig_factory or to null, or to conditionally do nothing if the C# factory is null)

Basically, because these similar but subtly different variables exist, it is harder for me to follow which method is being called, and hard for me to say when they are equal or different.

app_check/src/FirebaseAppCheck.cs
```diff
@@ -79,8 +87,18 @@ public sealed class FirebaseAppCheck {
   ///
   /// This method should be called before initializing the Firebase App.
   public static void SetAppCheckProviderFactory(IAppCheckProviderFactory factory) {
+    if (appCheckFactory == factory) return;
+
     appCheckFactory = factory;
-    // TODO(amaurice): Clear the provider map when the factory changes?
+    // Clear out the Providers that were previously made
```
AlmostMatt, line 93: Just extend this comment a bit to explain why the map is being cleared.

// Clear out the Providers that were previously made. When GetToken fails to find a provider in this map it will use appCheckFactory to create a new provider.

app_check/src/FirebaseAppCheck.cs
```diff
@@ -79,8 +87,18 @@ public sealed class FirebaseAppCheck {
...
+    providerMap.Clear();
+
+    // Register the callback for C++ SDK to use that will reach this factory.
```
AlmostMatt, line 96: I wonder a bit about why we need to conditionally set and clear getTokenDelegate when we set and clear the C# factory.

app_check/src/swig/app_check.i
AlmostMatt: Or, even more direct, is it possible to call GetTokenFromCSharpMethod by name instead of having a variable like g_get_token_from_csharp?

It seems like we always want GetToken to call GetTokenFromCSharpMethod, so I wonder what is the point of the variable that decides which C# method to call.

a-maurice: You've hit on the problem with your second comment.  As far as I know, there isn't a way for C++ to just be aware of the existence of the C# functions.  So instead we have C# call into C++, passing in function pointers to the C# functions, for C++ to then call into.

app_check/src/AppCheckToken.cs
```diff
@@ -28,6 +30,24 @@ public struct AppCheckToken {
 
   /// The time at which the token will expire.
   public System.DateTime ExpireTime { get; set; }
```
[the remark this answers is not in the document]
a-maurice: Done

app_check/src/FirebaseAppCheck.cs
a-maurice: Hmm, I'll keep an eye out for that.

app_check/src/swig/app_check.i
a-maurice: Yeah, because the C++/C# layer can be a bit tricky, especially with the memory management, it is following that pattern of setting a single generic factory, which passes everything to C# to then figure out.  I updated the comments to better reflect that, let me know if that helps clarify things.

app_check/src/FirebaseAppCheck.cs
a-maurice: Done

app_check/src/FirebaseAppCheck.cs
a-maurice: It isn't necessarily something we have to do, but just doing it for convenience of reducing the back and forth of C++ / Unity.  By setting it to null, the C++ part of the implementation can know that the C# side was cleared out, and pass errors back without having to reach up to C# first, which also involves putting things on the update queue.

app_check/src/swig/app_check.i
```diff
@@ -20,14 +20,195 @@
...
+  app_check_token.expire_time_millis = expire_ms;
+  callback(app_check_token, error_code, error_message);
+}
+
+// Wrapper that calls g_get_token_from_csharp, to get a new AppCheckToken
+// from the C# implementation. Logic to determine the provider is done on
+// the C# side.
+// Should be used with the callback logic to guarantee it is on the Unity thread.
+static void CallGetTokenFromCSharp(int key, const char* name) {
+  if (g_get_token_from_csharp) {
```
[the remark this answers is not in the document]
a-maurice: Added comments on both where they are defined, and called.

PR #678, opened 2023-04-19
Add App Check built in providers

a-maurice:
Add hooks into the built in App Check providers, by calling into the C++ implementations.
***

Testing locally
***

### Type of Change
***


app_check/src/DebugProvider.cs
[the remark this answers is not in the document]
a-maurice: Whoops, good catch

app_check/src/BuiltInProviderWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+    return tcs.Task;
+  }
+
+  private static void CompleteBuiltInGetTokenMethod(int key, System.IntPtr tokenCPtr,
+                                                    int error, string errorMessage) {
+    TaskCompletionSource<AppCheckToken> tcs;
+    if (s_pendingGetTokens.TryGetValue(key, out tcs)) {
+      s_pendingGetTokens.Remove(key);
+      if (error == 0) {
+        AppCheckTokenInternal tokenInternal = new AppCheckTokenInternal(tokenCPtr, false);
```
AlmostMatt, line 65: I'm not really sure how this line works, as I'm not sure what the swig-generated TokenInternal cosntructor looks like.
It is pretty clear what it is trying to do though, so as long as the testapp works things are ok.

app_check/src/BuiltInProviderWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+    if (providerInternal == null ||
+        AppCheckProviderInternal.getCPtr(providerInternal).Handle == System.IntPtr.Zero) {
+      throw new System.NullReferenceException();
+    }
+  }
+
+  public System.Threading.Tasks.Task<AppCheckToken> GetTokenAsync() {
+    int key = s_pendingGetTokenKey++;
+    TaskCompletionSource<AppCheckToken> tcs = new TaskCompletionSource<AppCheckToken>();
+    s_pendingGetTokens[key] = tcs;
```
AlmostMatt, line 54: Does anything need to be done to make access to s_pendingGetTokens threadsafe?

For example, if two firebase apps are in use simultaneously, they might both fetch app check tokens at once and they might both be adding and removing entries in this dictionary in parallel.

app_check/src/BuiltInProviderWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+    AppCheckUtil.GetTokenFromBuiltInProvider(providerInternal, key);
```
AlmostMatt, line 55: Call ThrowIfNull() before this line that depends on providerInternal.

At least, I think that is the pattern, to verify that the C++ type is not null before depending on it.
Alternatively complete the task early with an error, though I think throwing an exception is fine.

AlmostMatt (commented):
Some relatively minor feedback.

AlmostMatt:
Also, does the TestApp depend on the DebugProvider? Will there be an additional PR for that?

app_check/src/BuiltInProviderWrapper.cs
a-maurice: Added a comment explaining it better

app_check/src/BuiltInProviderWrapper.cs
a-maurice: Generally it should be around the Unity main thread, but added a lock around it, just to be safe.

app_check/src/BuiltInProviderWrapper.cs
a-maurice: Done

app_check/src/BuiltInProviderWrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+  private static void CompleteBuiltInGetTokenMethod(int key, System.IntPtr tokenCPtr,
+                                                    int error, string errorMessage) {
+    TaskCompletionSource<AppCheckToken> tcs;
+    lock (s_pendingGetTokens) {
+      if (!s_pendingGetTokens.TryGetValue(key, out tcs)) {
+        return;
+      }
+    }
+
+    s_pendingGetTokens.Remove(key);
```
AlmostMatt, line 75: I think this line should be part of the locked block as well

AlmostMatt (approved):

PR #706, opened 2023-05-05
Update RemoteConfig testapp to use the new Realtime methods

AlmostMatt:
### Description
Add GUI buttons to enable and disable auto-fetch by adding and removing config a update listener.
Add an automated test to verify that the config update listener is called with the correct keys.
Update an existing test to assert the fetched values.
***
### Testing
Ran testapp on a local android device and local ios simulator.
***

### Type of Change
***

### DIFF
--- a/remote_config/testapp/Assets/Firebase/Sample/RemoteConfig/UIHandlerAutomated.cs
+++ b/remote_config/testapp/Assets/Firebase/Sample/RemoteConfig/UIHandlerAutomated.cs
@@ -18,8 +18,8 @@ namespace Firebase.Sample.RemoteConfig {
 // Skip the Realtime RC test on desktop as it is not yet supported.
 #if (UNITY_IOS || UNITY_TVOS || UNITY_ANDROID) && !UNITY_EDITOR
         TestAddOnConfigUpdateListener,
-#endif  // !(UNITY_IOS || UNITY_TVOS || UNITY_ANDROID) || UNITY_EDITOR
         TestAddAndRemoveConfigUpdateListener,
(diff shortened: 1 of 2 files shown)


remote_config/testapp/Assets/Firebase/Sample/RemoteConfig/UIHandlerAutomated.cs
```diff
@@ -13,6 +15,11 @@ public class UIHandlerAutomated : UIHandler {
       Func<Task>[] tests = {
         TestDisplayData,
         TestDisplayAllKeys,
+// Skip the Realtime RC test on desktop as it is not yet supported.
+#if (UNITY_IOS || UNITY_TVOS || UNITY_ANDROID) && !UNITY_EDITOR
+        TestAddOnConfigUpdateListener,
+#endif  // !(UNITY_IOS || UNITY_TVOS || UNITY_ANDROID) || UNITY_EDITOR
+        TestAddAndRemoveConfigUpdateListener,
```
cynthiajoan, line 22: this line should also be inside the #if right?

cynthiajoan (approved):

remote_config/testapp/Assets/Firebase/Sample/RemoteConfig/UIHandlerAutomated.cs
AlmostMatt: Yea it probably should be. It technically passed on desktop since add and remove are no-op on desktop, but might as well not test those methods.

PR #723, opened 2023-05-13
Remove more accidentally exposed methods

chkuang-g:
- Remove `FirebaseAuth.current_user_DEPRECATED`
- Add a universal rule to ignore any method ended with `LastResult_DEPRECATED`
- Move `%ignore provider_data_DEPRECATE`

***

GHA
***

### Type of Change
***

### DIFF
--- a/auth/src/swig/auth.i
+++ b/auth/src/swig/auth.i
@@ -85,6 +85,7 @@ namespace auth {
 %ignore User::GetTokenThreadSafe;
 %ignore User::GetTokenLastResult;
 %ignore User::provider_data;
+%ignore User::provider_data_DEPRECATED;
 %ignore User::is_email_verified;
 %ignore User::is_anonymous;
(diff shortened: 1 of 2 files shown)


jonsimantov (approved):

auth/src/swig/auth.i
```diff
@@ -85,6 +85,7 @@ namespace auth {
 %ignore User::GetTokenThreadSafe;
 %ignore User::GetTokenLastResult;
 %ignore User::provider_data;
+%ignore User::provider_data_DEPRECATED;
```
AlmostMatt, line 88: provider_data_DEPRECATED is the field for which I made a change previously. I think that User:: is the wrong namespace to ignore it.

chkuang-g: Uh this is weird. This section is namespaced and you are correct: it does not work here.

Changed it back

PR #748, opened 2023-05-31
Simplify Messaging iOS tests on CI

a-maurice:
Messaging on iOS prompts the user for permissions on start, which causes problems with CI.  For now, disable those tests when running from CI.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/5138723229
***

### Type of Change
***


messaging/testapp/Assets/Firebase/Sample/Messaging/UIHandlerAutomated.cs
```diff
@@ -13,7 +13,7 @@ public class UIHandlerAutomated : UIHandler {
 
     private const string TestTopic = "TestTopic";
     private const string ServerKey = "REPLACE_WITH_YOUR_SERVER_KEY";
-    private const string FirebaseBackendUrl = "REPLACE_WITH_YOUR_BACKEND_URL";
+    private const string FirebaseBackendUrl = "https://fcm.googleapis.com/fcm/send";
```
sunmou99, line 16: Is this change intentional?

a-maurice: Yeah, there is no reason to obfuscate this url, it is publicly known.

scripts/gha/build_testapps.py
```diff
@@ -426,6 +426,12 @@ def setup_unity_project(dir_helper, setup_options):
   _copy_unity_assets(dir_helper, setup_options.testapp_file_filters)
   _add_menu_scene(dir_helper)
   _add_automated_test_runner(dir_helper)
+  # If running on CI, copy over the file that defines that custom scripting symbol
+  # For more info, see: https://docs.unity3d.com/Manual/CustomScriptingSymbols.html
```
sunmou99, line 430: It is also possible to use [PlayerSettings.SetScriptingDefineSymbolsForGroup](https://docs.unity3d.com/ScriptReference/PlayerSettings.SetScriptingDefineSymbolsForGroup.html) .
This is what we did in the past: https://github.com/firebase/firebase-unity-sdk/blob/main/scripts/gha/integration_testing/AppBuilderHelper.cs
But you can decide how to implement it.

sunmou99 (approved):

scripts/gha/build_testapps.py
a-maurice: I guess for now we can submit this, and in the future if we have the need for more, or this does have some issue, we can look into the other method.

PR #769, opened 2023-06-22
Mila/dynamically set target backend in test

milaGGL:
Update Firestore test helpers to dynamically set target backend: production, emulator.
Set Firestore emulator as default backend, run tests against production backend only on mobile devices (iOS, tvOS, Android).
Add Firestore emulator to CI test work flow.


scripts/gha/build_testapps.py
```diff
@@ -556,8 +556,7 @@ def patch_android_env(unity_version):
     # but currently the GitHub runners have 33 as their max.
     logging.info("Uninstall Android platform android-33")
     _run([sdkmanager_path, "--uninstall",
-          "platforms;android-33", "platforms;android-33-ext4", "platforms;android-33-ext5",
-          "platforms;android-34"], check=False)
+          "platforms;android-33", "platforms;android-33-ext4", "platforms;android-33-ext5"], check=False)
```
milaGGL, line 559: reverted back the changes merged from master. TODO: merge master into this pr.

firestore/testapp/Assets/Firebase/Sample/Firestore/UIHandler.cs
```diff
@@ -60,6 +60,34 @@ public class UIHandler : MonoBehaviour {
...
+    }
+
+    private Backend TargetBackend = GetTargetBackend();
+    // For local testing purpose, manually modify the port number if Firestore
+    // emulator is running at a different port.
+    private string EmulatorPort = "8080";
+
+    private static Backend GetTargetBackend() {
+      // Set Firestore emulator as default backend.
+      Backend targetBackend = Backend.Emulator;
```
a-maurice, line 76: I think it would make more sense to use Production as the default, since additional steps are needed to set up the Emulator

tom-andersen: I am a little confused by changes. It appears you are starting emulator as part of CI now. Does that mean we are NOT testing against prod anymore?

milaGGL: Discussed offline. CI tests on UNITY_EDITOR (playmode) or UNITY_STANDALONE(macOS, Linux, Windows) are default to run against Emulator. And other platforms(iOS, tvOS, Android) will run against prod.

a-maurice: My major concern is that someone will try to run the Firestore testapp locally in editor without knowing that they have to also spin up an Emulator, and then get a bunch of errors.  But if you all are fine with that though, that's fine.

a-maurice (approved):

firestore/testapp/Assets/Firebase/Sample/Firestore/UIHandler.cs
milaGGL: While running Firestore tests against the emulator, it will print a line of "Running tests against Firestore Emulator.". Hopefully, this would give the users some hint.

PR #770, opened 2023-06-23
Uninstall android-34 on the GHA machines

a-maurice:
The GHA machines added android-34, which has problems with the Unity Android build when using proguard.
***

***

### Type of Change
***


chkuang-g (approved):
Not very sustainable. We should invest in a generic solution for this.

AlmostMatt:
> Not very sustainable. We should invest in a generic solution for this.

I have a more generalized solution in https://github.com/firebase/firebase-unity-sdk/pull/771

PR #771, opened 2023-06-23
Uninstall all versions of android platform 33 or newer

AlmostMatt:
This generalizes some previous changes (#604) that uninstalled android platform 33 to also uninstall newer versions.
Recent GitHub runners have had android platform 34 installed.

This also changes from list_installed to list as when I tested locally, list_installed was not a recognized command. List outputs installed and available.
***
### Testing
Build with this change
https://github.com/firebase/firebase-unity-sdk/actions/runs/5360532633
***

### Type of Change
***

### DIFF
--- a/scripts/gha/build_testapps.py
+++ b/scripts/gha/build_testapps.py
@@ -89,6 +89,7 @@ import stat
 import shutil
 import subprocess
 import time
+import re
 import requests
 import zipfile
 import json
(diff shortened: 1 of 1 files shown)


scripts/gha/build_testapps.py
```diff
@@ -550,21 +551,25 @@ def patch_android_env(unity_version):
...
-    logging.exception("Failed to uninstall Android platform android-33")
-
+    logging.exception("Failed to uninstall Android platforms;android-33")
   try:
-    # List the installed packages to make it easier to notice if any incompatible packages are present.
+    # List the installed packages to make it easier to debug if any incompatible
+    # packages are present.
     logging.info("Listing installed sdks")
-    _run([sdkmanager_path, "--list_installed"], check=False)
+    _run([sdkmanager_path, "--list"], check=False)
```
a-maurice, line 572: What is the difference between list and list_installed?

AlmostMatt: My understanding is that list_installed is only an option for the windows sdkmanager.bat, but list is supported by both mac and windows sdkmanagers. This page mentions 'list' but has no mention of 'list_installed':
https://developer.android.com/tools/sdkmanager#list
I switched to 'list' as it should work even if we switched the android build to happen on mac or linux instead of windows.

list_installed outputs a list of installed packages. list outputs the same list of installed packages followed by an additional list of 'available' packages. This does mean that it can result in 'false positives' where it will attempt to uninstall an available package, but I verified that the uninstall command succeeds even if it attempts to uninstall something that was not already installed.

You can see an example of the 'list' output in the 'Build integration tests' step of this run:
https://github.com/firebase/firebase-unity-sdk/actions/runs/5360532633/jobs/9725527898

a-maurice: Cool, as long as there is some way to look at the logs and see what is installed vs not, that sounds good.

a-maurice (approved):

PR #774, opened 2023-06-28
Improve the iOS simulator install logic

a-maurice:
Add logic to check which versions of iOS/tvOS are available, and fallback to the latest if the requested one isn't available. Also change to using xcodes to install iOS simulators, since xcode-install has been deprecated in favor of it.
***

Running locally, and https://github.com/firebase/firebase-unity-sdk/actions/runs/5395850671
***

### Type of Change
***


sunmou99 (approved):
LGTM

Issue #569, opened 2022-12-12
[Bug] Firestore Android using too many global refs


a-maurice:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2021.3.13f1
  * Firebase Unity SDK version: 10.2.0
  * Source you installed the SDK: unitypackage
  * Problematic Firebase Component: Firestore
  * Other Firebase Components in use: Crashlytics
  * Additional SDKs you are using: N/A
  * Platform you are using the Unity editor on: All
  * Platform you are targeting: Android
  * Scripting Runtime: N/A

### [REQUIRED] Please describe the issue here:

(Please list the full steps to reproduce the issue. Include device logs, Unity logs, and stack traces if available.)

#### Steps to reproduce:

Have you been able to reproduce this issue with just the Firebase Unity quickstarts (this GitHub project)? No
What's the issue repro rate? (eg 100%, 1/5 etc) 100%

What happened? How can we make the problem occur?

As identified in https://github.com/firebase/quickstart-unity/issues/1193, there is a problem with large sets of data with Firestore on Android, where it can use too many global refs, hitting the cap and throwing a JNI error.  An initial fix was added with the 10.1.0 release which reduces the number of global refs used, but it can still potentially be hit.

Googlers see b/251869890 for more details.

dconeybe:
FYI: If you see an error like this in the Android logcat

```
JNI ERROR (app bug): global reference table overflow (max=51200)
```

then you have hit this issue.

dconeybe:
Also, if your stack trace includes methods like `art::JavaVMExt::AddGlobalRef(...)` or `art::JNI::NewGlobalRef(...)` then that is another strong indicator that it is this issue that you are experiencing (e.g. https://github.com/firebase/firebase-unity-sdk/issues/586).

dconeybe:
Update: The fix, unfortunately, is not yet ready and did not make it into last week's v10.4.0 release. We ran into some technical issues with our solution and need to do some refactoring. We hope to have the fix ready for the next release, which is scheduled for mid February 2023. If interested, you can follow the details of the fix in https://github.com/firebase/firebase-cpp-sdk/pull/1176.

dconeybe:
Update: The fix is still being finalized. We keep running into unexpected technical issues with our solution and also some competing priorities. We will update this issue when there is a build to test out.

nitzanwilnai:
Hi, did the fix make it into 10.5.0 ?

[...]

PR #781, opened 2023-07-04
Firestore: Add a test and readme about the android global refs crash

dconeybe:
Add a test and release notes entry for a fix to an intermittent crash in Firestore on Android when large documents and/or a large number of documents with large arrays or maps were used. See https://github.com/firebase/firebase-unity-sdk/issues/569 and https://github.com/firebase/firebase-cpp-sdk/pull/1364 for details.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -71,6 +71,13 @@ Support
 
 Release Notes
 -------------
+### Upcoming Release
+- Changes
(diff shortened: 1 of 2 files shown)


dconeybe:
Integration tests: https://github.com/firebase/firebase-unity-sdk/actions/runs/5459341301

wu-hui (approved):

PR #829, opened 2023-08-10
Refactor FirebaseUser to handle C++ objects better

a-maurice:
The C++ object for FirebaseUser isn't reliable enough in memory for the C# object to hold onto a reference to it (for example, if the AuthResult is deleted, the User will be as well).  Some users were seeing crashes being caused by this.
Fix it by fetching the C++ object when needed, which guarantees that it will be a valid C++ object.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/5863182504
***

### Type of Change
***

### DIFF
--- a/auth/src/FirebaseUser.cs
+++ b/auth/src/FirebaseUser.cs
@@ -1,0 +1,363 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ using System.Threading.Tasks;
+
+namespace Firebase.Auth {
+
+/// @brief Firebase user account object.
+///
+/// This class allows you to manipulate the profile of a user, link to and
+/// unlink from authentication providers, and refresh authentication tokens.
+public sealed class FirebaseUser : UserInfoInterface {
+  private FirebaseAuth authProxy;
+
+  internal FirebaseUser(FirebaseAuth auth) {
+    authProxy = auth;
+  }
+
+  private FirebaseUserInternal GetValidFirebaseUserInternal() {
+    if (authProxy == null) {
+      throw new System.NullReferenceException();
+    } else {
+      FirebaseUserInternal userInternal = authProxy.CurrentUserInternal;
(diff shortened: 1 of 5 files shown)


auth/src/swig/auth.i
```diff
@@ -1193,21 +1118,7 @@ static CppInstanceManager<Auth> g_auth_instances;
 
 %typemap(csclassmodifiers) firebase::auth::Auth "public sealed class";
 %typemap(csclassmodifiers) firebase::auth::User "public sealed class";
```
chkuang-g, line 1120: I believe `firebase::auth::User` should be `internal sealed class` now since we do not want to expose `FirebaseUserInternal`

auth/src/swig/auth.i
```diff
@@ -1313,11 +1225,14 @@ static CppInstanceManager<Auth> g_auth_instances;
...
   /// The currently signed-in @ref FirebaseUser, or null if there isn't one.
   public FirebaseUser User {
-    // Both FirebaseAuth.CurrentUser and AuthResult.User returns null if the
-    // user is invalid.
-    get { return (UserInternal != null && UserInternal.IsValid()) ? UserInternal : null; }
+    // Return the Auth's User, since there can currently only be one user per Auth.
+    get {
+      return authProxy != null ? authProxy.CurrentUser : null;
+    }
   }
```
chkuang-g, line 1236: I am expecting `SignInResult.User` would have exactly the same issue. I think we should properly fix it until we actually remove this API.

chkuang-g (changes requested):

auth/src/FirebaseUser.cs
```diff
@@ -0,0 +1,407 @@
...
+  /// @ref ReauthenticateWithProviderAsync(FederatedAuthProvider) instead.
+  ///
+  /// Reauthenticate a user via a federated auth provider.
+  ///
+  /// @note: This operation is supported only on iOS, tvOS and Android
+  /// platforms. On other platforms this method will return a Future with a
+  /// preset error code: kAuthErrorUnimplemented.
+  [System.Obsolete("Please use `Task<AuthResult> ReauthenticateWithProviderAsync(FederatedAuthProvider)` instead", false)]
+  public async Task<SignInResult> ReauthenticateWithProviderAsync_DEPRECATED(FederatedAuthProvider provider) {
+    FirebaseUserInternal userInternal = GetValidFirebaseUserInternal();
```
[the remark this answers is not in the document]
a-maurice: It is using it, unless I'm misunderstanding what you mean.

auth/src/swig/auth.i
a-maurice: Ah, it was getting changed to internal by the post process thing that changes things with "Internal" in the name, but it is definitely confusing, so will change it here as well.

auth/src/swig/auth.i
a-maurice: Agreed.  SignInResult already does this logic, of using the authProxy and returning the User based on it instead.

chkuang-g (approved):

PR #848, opened 2023-08-30
Change how Messaging testapp gets registration token

a-maurice:
The OnTokenReceived callback is only called for new Messaging Tokens, so if the App has a cached token, it would not receive a new token, and thus the tests would timeout while waiting for it.  Change the tests to use GetTokenAsync if necessary to get a token, and throw exceptions if the token is still missing.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/6028673810
***

### Type of Change
***


messaging/testapp/Assets/Firebase/Sample/Messaging/UIHandlerAutomated.cs
```diff
@@ -42,6 +42,8 @@ public class UIHandlerAutomated : UIHandler {
     }
 
     protected override void Start() {
+      Firebase.FirebaseApp.LogLevel = Firebase.LogLevel.Verbose;
```
cynthiajoan, line 45: revert this?

cynthiajoan (approved):

PR #876, opened 2023-10-05
Add retries to the FTL integration test runs

a-maurice:
Add retry attempts to the FTL integration tests running on GHA, to work around device flakiness.  Adds a custom validator, since the FTL run reports success regardless of if the tests contained succeeded or failed.
***

Run with intentional failure to see the retries: https://github.com/firebase/firebase-unity-sdk/actions/runs/6475956332
Normal run: https://github.com/firebase/firebase-unity-sdk/actions/runs/6476383922
***

### Type of Change
***


sunmou99 (approved):
LGTM.
Sounds like that there are cases when unit tests failed but FTL return success.

PR #937, opened 2024-02-05
Update dependencies to latest, and other fixes

a-maurice:
Update to the latest C++ SDK, and the equivalent Android/iOS dependencies.  Update the versions of python and Node used by the GHA runners, because the CLI needs it.  Increase the timeout range on the macOS build, since the build seems to be taking longer because of dependency changes.
***

Runs on another branch with these changes
***

### Type of Change
***

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -71,6 +71,12 @@ Support
 
 Release Notes
 -------------
+### Upcoming
+- Changes
+    - General: Update to Firebase C++ SDK version 11.8.0.
+    - General (Android): Update to Firebase Android BoM version 32.7.1.
+    - General (iOS): Update to Firebase Cocoapods version 10.20.0.
+
 ### 11.6.0
 - Changes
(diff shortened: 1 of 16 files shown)


docs/readme.md
```diff
@@ -71,6 +71,12 @@ Support
 
 Release Notes
 -------------
+### Upcoming
```
cynthiajoan, line 74: still upcoming?

cynthiajoan (approved):

docs/readme.md
a-maurice: Yeah, not updating the version number of the Unity SDK yet, so upcoming (though hopefully not too much longer)

rubenwe: Looking forward to it! We really, really need the updated versions. Thanks!

a-maurice: Sorry for the delays, we are still fixing a problem with the Android dependencies, which we are still testing and iterating against different Unity editor versions, but again, hopefully will be available soon.

PR #1011, opened 2024-05-01
Fix formatting exception for unknown property

nbadal:
When an annotated class is set to `UnknownPropertyHandling.Throw`, an unexpected exception is thrown rather than the intended readable one. `string.Format()` has an invalid format string, as it expects exclusively argument indexes.
 
***

Tested by running the same scenario with the patch, and the expected exception is thrown to the calling code.
***

### Type of Change
***


a-maurice (approved):

a-maurice:
Thanks for the fix!

PR #1028, opened 2024-05-14
Retry test failures on Unity

jonsimantov:
It will retry all **build** errors that match timeout/network error regexes, but not any compiler errors or other issues. (If there are other instances in which the builds sometimes fail transiently, we can add that.)

Not sure under what conditions to actually trigger the **test** retry, so it does it unconditionally.

***

Tested manually with the "scheduled_trigger" check in integration_tests.yaml commented out, and failure forced.

This run: https://github.com/firebase/firebase-unity-sdk/actions/runs/9087398748
***

### Type of Change
***

### DIFF
--- a/scripts/gha/retry_test_failures.py
+++ b/scripts/gha/retry_test_failures.py
@@ -1,0 +1,128 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


scripts/gha/retry_test_failures.py
```diff
@@ -0,0 +1,128 @@
+# Copyright 2023 Google LLC
```
cynthiajoan, line 1: 2024

jonsimantov: Eh, this is mostly a copy of the script from C++ that was written in 2023, should it still be 2024?

cynthiajoan (approved):

PR #1068, opened 2024-07-16
Rewrite Messaging Swig logic to prefer handwritten public files

a-maurice:
Rewrite the Messaging Swig logic, so that the public API is in handwritten C# files, and the generated code is instead internal logic to be called.  This fixes a handful of missing docs generation issues, and deprecates the Dispose/Disposing calls, since those are no longer necessary to expose at all.

***

Confirmed the docs generation looks correct by hand
https://github.com/firebase/firebase-unity-sdk/actions/runs/9964644977
***

### Type of Change
***

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -74,7 +74,7 @@ Release Notes
 ### Upcoming release
 - Changes
     - Messaging: Deprecated the Dispose functions, as they are no longer
-      necessary to use to clean up memory.
+      necessary for cleaning up memory.
 
 ### 12.1.0
 - Changes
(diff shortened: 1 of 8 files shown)


docs/readme.md
```diff
@@ -71,6 +71,11 @@ Support
 
 Release Notes
 -------------
+### Upcoming release
+- Changes
+    - Messaging: Deprecated the Dispose functions, as they are no longer
+      necessary to use to clean up memory.
```
jonsimantov, line 77: nit: "necessary for cleaning up memory"

jonsimantov (approved):

Issue #1091, opened 2024-08-21
[Bug] Crash when opening application by clicking push notifications


EllieTellie:
### Description

The application crashes when tapping a (remote) push notification on iOS & Android when using Firebase Messaging v12.2.0.
We have a splash screen and it briefly shows that before crashing within 1-2 seconds.
This seems to be a regression as our old version 11.7.0 this crash does not occur.

Crash rate is 100% in our scenario.

The crash stack trace references deconstructors and is called from the gargage collector.
We do have incremental garbage collection enabled in our Unity project.

Here's the relevant stack trace
0   libsystem_kernel.dylib        	0x00000001eee4342c __pthread_kill + 8 (:-1)
1   libsystem_pthread.dylib       	0x0000000202be2c0c pthread_kill + 268 (pthread.c:1721)
2   libsystem_c.dylib             	0x00000001adce6c34 __abort + 136 (abort.c:159)
3   libsystem_c.dylib             	0x00000001adce6bac abort + 192 (abort.c:126)
4   libsystem_malloc.dylib        	0x00000001b5f24588 malloc_vreport + 896 (malloc_printf.c:251)
5   libsystem_malloc.dylib        	0x00000001b5f241f8 malloc_report + 64 (malloc_printf.c:290)
6   libsystem_malloc.dylib        	0x00000001b5f237b0 find_zone_and_free + 528 (malloc.c:2793)
7   UnityFramework                	0x000000010ec30f4c operator delete(void*) + 48 (MemoryManager.cpp:275)
8   UnityFramework                	0x000000010fc28d44 firebase::messaging::Notification::~Notification() + 72
9   UnityFramework                	0x000000010fc23da4 firebase::messaging::Notification::~Notification() + 28
10  UnityFramework                	0x000000010fc29370 firebase::messaging::Message::~Message() + 64
11  UnityFramework                	0x000000010fc23dd0 firebase::messaging::Message::~Message() + 28
12  UnityFramework                	0x000000010fc2cbe4 Firebase_Messaging_CSharp_delete_FirebaseMessageInternal + 104
13  UnityFramework                	0x00000001109c7978 FirebaseMessagingInternalPINVOKE_delete_FirebaseMessageInternal_mD15EF62AC59E1DA718AEDD7AFE2D4AACD824A304 + 8 (Firebase.Messaging.cpp:6042)
[...]

PR #1101, opened 2024-09-03
Add null checks, and improve swig error handling

a-maurice:
Add missing null checks to new Messaging conversion logic, and improve the swig conversion logic to better handle if an exception is thrown, to prevent a double free.  This was causing a crash when opening a notification.
***

Running locally.
GHA: https://github.com/firebase/firebase-unity-sdk/actions/runs/10691311630
***

### Type of Change
***

https://github.com/firebase/firebase-unity-sdk/issues/1091

### DIFF
--- a/messaging/src/swig/messaging.i
+++ b/messaging/src/swig/messaging.i
@@ -69,7 +69,7 @@ public:
   // OnMessage() callback in this class.  SWIGSTDCALL is used here as C#
   // delegates *must* be called using the stdcall calling convention rather
   // than whatever the compiler defines.
-  typedef int (SWIGSTDCALL *MessageReceivedCallback)(void* message);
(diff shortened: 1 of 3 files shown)


messaging/src/swig/messaging.i
```diff
@@ -316,18 +316,22 @@ void SendPendingEvents() {
     // messageReceivedDelegate.
     [MonoPInvokeCallback(typeof(MessageReceivedDelegate))]
     private static int MessageReceivedDelegateMethod(System.IntPtr message) {
+      int tookOwnership = 0;
```
jonsimantov, line 319: If this has to be an int vs. a bool, add a comment for why?

jonsimantov (approved):

PR #1108, opened 2024-09-10
Refactor Analytics swig logic

a-maurice:
Refactor the Analytics swig logic to use more handwritten public APIs.  Includes a script to generate the predefined Analytics variables directly, instead of having to access the C++ library for each one.  Deprecates Dispose logic, since it is no longer necessary, as the C++ objects aren't being exposed to C# at all now.
***

Running locally, verifying events appear in the debug view of the console.

Just documentation: https://github.com/firebase/firebase-unity-sdk/actions/runs/11372138958
SDK build: https://github.com/firebase/firebase-unity-sdk/actions/runs/11371225430
***

### Type of Change
***


analytics/generate_constants.py
```diff
@@ -0,0 +1,83 @@
...
+  ('static const char\*const k', 'public static string ')
+]
+
+
+def main(unused_argv):
+  """Convert the C++ file into C#, going line-by-line to edit it."""
+  with open(FLAGS.cpp_header) as input_file:
+    with open(FLAGS.csharp_file, 'w') as output_file:
+      # Write the initial lines at the top
+      output_file.write('// Copyright %s Google Inc. All Rights Reserved.\n\n' %
```
jonsimantov, line 51: I think technically this is supposed to be Google LLC now.

jonsimantov (approved):

analytics/generate_constants.py
a-maurice: Fixed

PR #1114, opened 2024-09-24
Update tools to use Unity 2021

a-maurice:
Update the build and test tools to use Unity 2021.3.43f1.  Update the testapps to use that version as well.  Update to use Java 11 instead of Java 8, needed by the latest version of C++ because of gradle changes.  Update the XCode capabilities to use the ProjectCapabilityManager instead of copying entitlement files and doing it manually.
***

Various GHA runs
***

### Type of Change
***


jonsimantov (approved):
Actually relatively straightforward, except the capabilities stuff, which is much nicer now than it was before.

PR #1192, opened 2025-02-25
VertexAI - Add more functionality around options in, and feedback out

a-maurice:
Fill in more support for the various input options, for example SafetySettings, and the feedback out, like PromptFeedback.
Adds an Internal namespace that includes extension methods to help with the Json conversions.
***

Running tests locally.
***

### Type of Change
***


vertexai/src/GenerativeModel.cs
```diff
@@ -183,7 +187,7 @@ private async Task<CountTokensResponse> CountTokensAsyncInternal(
   }
 
   private string GetURL() {
-    return "https://firebaseml.googleapis.com/v2beta" +
+    return "https://firebasevertexai.googleapis.com/v1beta" +
```
DellaBitta, line 190: Shouldn't this be a constant at the top of the file somehow?

DellaBitta (approved):

vertexai/src/GenerativeModel.cs
a-maurice: Maybe, but since it is only used here, and also it seems like this logic is going to have to be rewritten in the nearish future anyway with different url layouts based on different scenarios, I'm just going to leave it as is for now.

PR #1197, opened 2025-02-28
VertexAI - Add Streaming responses, and some tests

a-maurice:
Add the logic for Streaming a response. Add tests for setting the various options, multiple candidates, and streaming.
***

https://github.com/firebase/firebase-unity-sdk/actions/runs/13579798146

https://github.com/firebase/firebase-unity-sdk/actions/runs/13580286421
***

### Type of Change
***

### DIFF
--- a/vertexai/src/GenerativeModel.cs
+++ b/vertexai/src/GenerativeModel.cs
@@ -142,11 +142,6 @@ public class GenerativeModel {
   }
 #endregion
 
-  private void SetRequestHeaders(HttpRequestMessage request) {
-    request.Headers.Add("x-goog-api-key", _firebaseApp.Options.ApiKey);
-    request.Headers.Add("x-goog-api-client", "genai-csharp/0.1.0");
-  }
-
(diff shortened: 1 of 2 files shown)


vertexai/src/GenerativeModel.cs
```diff
@@ -148,17 +142,20 @@ public Chat StartChat(IEnumerable<ModelContent> history) {
   }
 #endregion
 
+  private void SetRequestHeaders(HttpRequestMessage request) {
+    request.Headers.Add("x-goog-api-key", _firebaseApp.Options.ApiKey);
+    request.Headers.Add("x-goog-api-client", "genai-csharp/0.1.0");
```
cynthiajoan, line 147: Is this fixed number?

cynthiajoan (approved):

vertexai/src/GenerativeModel.cs
a-maurice: Good catch, it shouldn't be.  Getting the actual version number will take some additional changes to App to expose it, so I'm going to mark it with a TODO for now.

PR #1230, opened 2025-04-15
[VertexAI] Add initial support for BiDi streaming

a-maurice:
Adds basic support for BiDirectional streaming support for VertexAI. Note that the feature is experimental still, and very likely to change.
***

Running locally.
***

### Type of Change
***

### DIFF
--- a/vertexai/src/LiveContentResponse.cs
+++ b/vertexai/src/LiveContentResponse.cs
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Google.MiniJSON;
+using Firebase.VertexAI.Internal;
+using System.Linq;
+using System;
+
+namespace Firebase.VertexAI {
+
+/// <summary>
+/// Represents the response from the model for live content updates.
+/// </summary>
+public readonly struct LiveContentResponse {
+  /// <summary>
+  /// Represents the status of the `LiveContentResponse`.
+  /// </summary>
+  public enum LiveResponseStatus {
+    /// <summary>
+    /// Indicates response has no special status associated with it.
+    /// </summary>
+    Normal,
+    /// <summary>
+    /// Indicates the model is done generating content.
+    /// </summary>
+    TurnComplete,
+    /// <summary>
+    /// Indicates a client message has interrupted the current generation.
+    /// </summary>
+    Interrupted
+  }
+
+  /// <summary>
+  /// The main content data of the response. This can be `null` if there is no content.
+  /// </summary>
+  public ModelContent? Content { get; }
+
+  /// <summary>
+  /// The status of the response.
+  /// </summary>
+  public LiveResponseStatus Status { get; }
+
+  private readonly ReadOnlyCollection<ModelContent.FunctionCallPart> _functionCalls;
+
+  /// <summary>
+  /// A list of [FunctionCallPart] included in the response, if any.
+  ///
+  /// This will be empty if no function calls are present.
+  /// </summary>
+  public IEnumerable<ModelContent.FunctionCallPart> FunctionCalls =>
+      _functionCalls ?? new ReadOnlyCollection<ModelContent.FunctionCallPart>(
+          new List<ModelContent.FunctionCallPart>());
+
+  // TODO: Add this
+  //public IEnumerable<string> FunctionIdsToCancel { get; }
+
+  /// <summary>
+  /// The response's content as text, if it exists.
+  /// </summary>
+  public string Text {
+    get {
+      string str = "";
+      if (Content != null) {
+        foreach (var part in Content?.Parts) {
+          if (part is ModelContent.TextPart textPart) {
+            str += textPart.Text;
+          }
+        }
+      }
+      return str;
+    }
+  }
+
+  /// <summary>
+  /// The response's content that was audio, if it exists.
+  /// </summary>
+  public IEnumerable<byte[]> Audio {
+    get {
+      return Content?.Parts
+          .OfType<ModelContent.InlineDataPart>()
+          .Where(part => part.MimeType == "audio/pcm")
+          .Select(part => part.Data.ToArray());
+    }
+  }
+
+  /// <summary>
+  /// The response's content that was audio, if it exists, converted into floats.
+  /// </summary>
+  public IEnumerable<float[]> AudioAsFloat {
+    get {
+      return Audio?.Select(ConvertBytesToFloat);
+    }
+  }
+
+  private float[] ConvertBytesToFloat(byte[] byteArray) {
+    // Assumes 16 bit encoding, which would be two bytes per sample.
+    int sampleCount = byteArray.Length / 2;
+    float[] floatArray = new float[sampleCount];
+
+    for (int i = 0; i < sampleCount; i++) {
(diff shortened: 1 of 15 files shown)


vertexai/src/LiveContentResponse.cs
```diff
@@ -0,0 +1,174 @@
...
+namespace Firebase.VertexAI {
+
+/// <summary>
+/// Represents the response from the model for live content updates.
+/// </summary>
+public readonly struct LiveContentResponse {
+  /// <summary>
+  /// Represents the status of the `LiveContentResponse`.
+  /// </summary>
+  public enum LiveResponseStatus {
```
cynthiajoan, line 33: since we agree to follow proto, do you want to make the update now or let's do that in a follow up?

vertexai/src/LiveGenerationConfig.cs
```diff
@@ -0,0 +1,208 @@
...
+    return dict;
+  }
+}
+
+/// <summary>
+/// A struct defining model parameters to be used when generating live session content.
+/// </summary>
+public readonly struct LiveGenerationConfig {
+  readonly SpeechConfig? _speechConfig;
+  readonly List<ResponseModality> _responseModalities;
```
cynthiajoan, line 66: why these two don't need to be private?

vertexai/src/LiveGenerationConfig.cs
```diff
@@ -0,0 +1,208 @@
...
+  /// > for more details.</param>
+  /// 
+  /// <param name="stopSequences">A set of up to 5 `String`s that will stop output generation. If specified,
+  /// the API will stop at the first appearance of a stop sequence. The stop sequence will not
+  /// be included as part of the response. See the
+  /// [Cloud documentation](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#generationconfig)
+  /// for more details.</param>
+  public LiveGenerationConfig(
+      SpeechConfig? speechConfig = null,
+      IEnumerable<ResponseModality> responseModalities = null,
```
cynthiajoan, line 165: This should also be optional? IEnumerable<ResponseModality>?

vertexai/src/LiveGenerativeModel.cs
```diff
@@ -0,0 +1,131 @@
...
+  ///
+  /// This function handles the WebSocket connection setup and returns an `LiveSession`
+  /// object that can be used to communicate with the service.
+  /// </summary>
+  /// <param name="cancellationToken">The token that can be used to cancel the creation of the session.</param>
+  /// <returns>The LiveSession, once it is established.</returns>
+  public async Task<LiveSession> ConnectAsync(CancellationToken cancellationToken = default) {
+    ClientWebSocket clientWebSocket = new();
+
+    string endpoint = $"wss://firebasevertexai.googleapis.com/ws/google.firebase.vertexai.v1beta.LlmBidiService/BidiGenerateContent/locations/{_location}?key={_firebaseApp.Options.ApiKey}";
```
cynthiajoan, line 82: 'firebasevertexai.googleapis.com' and 'v1beta' might worth to be an attribute that could change in the future.

vertexai/src/LiveGenerativeModel.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    // Set initial headers
+    clientWebSocket.Options.SetRequestHeader("x-goog-api-client", "genai-csharp/0.1.0");
```
cynthiajoan, line 85: '0.1.0' be runtime version?

vertexai/src/ModelContent.cs
```diff
@@ -311,21 +311,36 @@ internal static ModelContent FromJson(Dictionary<string, object> jsonDict) {
       jsonDict.ParseObjectList("parts", PartFromJson, JsonParseOptions.ThrowEverything));
   }
 
-  private static FunctionCallPart FunctionCallPartFromJson(Dictionary<string, object> jsonDict) {
-    return new FunctionCallPart(
-      jsonDict.ParseValue<string>("name", JsonParseOptions.ThrowEverything),
-      jsonDict.ParseValue<Dictionary<string, object>>("args", JsonParseOptions.ThrowEverything));
+  private static InlineDataPart InlineDataPartFromJson(Dictionary<string, object> jsonDict) {
```
cynthiajoan, line 314: Why this is modifying the FunctionCallPart? I think we would need both.

vertexai/src/VertexAI.cs
```diff
@@ -106,6 +106,32 @@ public GenerativeModel GetGenerativeModel(
...
         toolConfig, systemInstruction, requestOptions);
   }
+
+  /// <summary>
+  /// Initializes a `LiveGenerativeModel` for real-time interaction.
+  /// 
+  /// - Note: Refer to [Gemini models](https://firebase.google.com/docs/vertex-ai/gemini-models) for
+  /// guidance on choosing an appropriate model for your use case.
+  /// </summary>
+  /// <param name="modelName">The name of the model to use, for example `"gemini-2.0-flash-live-preview-04-09"`; see
```
cynthiajoan, line 116: we should have a more stable live model name now.

cynthiajoan (approved):

vertexai/src/LiveGenerationConfig.cs
a-maurice: Whoops, they should be, good catch.

vertexai/src/LiveGenerationConfig.cs
a-maurice: The joys of C#, IEnumerable is nullable by default, hence no ? needed.

vertexai/src/LiveGenerativeModel.cs
a-maurice: I'll break it into a helper function for now, so that it can be easier to modify later once we know more.

vertexai/src/LiveGenerativeModel.cs
a-maurice: I copied the TODO I have from GenerativeModel. Basically, need to expose some info from C++ for this.

vertexai/src/ModelContent.cs
a-maurice: I moved it below, into an internal ModelContentJsonParsers class, so that LiveContentResponse could reuse the parsing logic.

vertexai/src/VertexAI.cs
a-maurice: Looking around, I couldn't find a great one. I'll switch it to "...flash-exp" for now, since that seems a bit more generic then the very long one with the date, and so might be a bit more stable.

a-maurice: After discussing it more, we will keep the longer name, since that is what the other docs have for now.

vertexai/src/LiveContentResponse.cs
a-maurice: Let's do it as a follow up, since we need to come to an agreement across all the platforms.

vertexai/src/LiveContentResponse.cs
```diff
@@ -0,0 +1,171 @@
...
+          turnComplete ? LiveResponseStatus.TurnComplete : LiveResponseStatus.Normal;
+      return new LiveContentResponse(
+        serverContent.ParseNullableObject("modelTurn", ModelContent.FromJson),
+        status);
+    } else if (jsonDict.TryParseValue("toolCall", out Dictionary<string, object> toolCall)) {
+      return new LiveContentResponse(
+          jsonDict.ParseObjectList("functionCalls", ModelContentJsonParsers.FunctionCallPartFromJson));
+    } else {
+      // TODO: Do we want to log this, or just ignore it?
+      return null;
```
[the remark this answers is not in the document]
a-maurice: Cleaned up slightly

vertexai/src/LiveContentResponse.cs
```diff
@@ -0,0 +1,171 @@
...
+  /// A list of [FunctionCallPart] included in the response, if any.
+  ///
+  /// This will be empty if no function calls are present.
+  /// </summary>
+  public IEnumerable<ModelContent.FunctionCallPart> FunctionCalls =>
+      _functionCalls ?? new ReadOnlyCollection<ModelContent.FunctionCallPart>(
+          new List<ModelContent.FunctionCallPart>());
+
+  // TODO: Add this
+  //public IEnumerable<string> FunctionIdsToCancel { get; }
```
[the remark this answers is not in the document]
a-maurice: Going to change how the LiveContentResponse is handled in a follow up change.

PR #1243, opened 2025-05-02
[FirebaseAI] Rename VertexAI to FirebaseAI

a-maurice:
Rename the product from VertexAI to FirebaseAI. Note that VertexAI is still an option as a backend, it is just the overall product that is being renamed.
***

Running the testapp locally
Build: https://github.com/firebase/firebase-unity-sdk/actions/runs/14805425451
Test: https://github.com/firebase/firebase-unity-sdk/actions/runs/14843992056
***

### Type of Change
***


firebaseai/src/Internal/FirebaseInterops.cs
```diff
@@ -56,7 +56,7 @@ static FirebaseInterops() {
   }
 
   private static void LogError(string message) {
-#if FIREBASE_VERTEXAI_DEBUG_LOGGING
+#if FIREBASEAI_DEBUG_LOGGING
     UnityEngine.Debug.LogError(message);
 #endif
```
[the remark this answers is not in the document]
a-maurice: Done

CMakeLists.txt
```diff
@@ -69,8 +69,8 @@ option(FIREBASE_INCLUDE_REMOTE_CONFIG
 option(FIREBASE_INCLUDE_STORAGE
        "Include the Cloud Storage for Firebase library."
        ${FIREBASE_INCLUDE_LIBRARY_DEFAULT})
-option(FIREBASE_INCLUDE_VERTEXAI
-       "Include the Vertex AI in Firebase library."
+option(FIREBASE_INCLUDE_FIREBASEAI
```
cynthiajoan, line 72: since we choose FIREBASE_INCLUDE_DYNAMIC_LINKS and  FIREBASE_INCLUDE_REMOTE_CONFIG, which makes me feel this should be FIREBASE_INCLUDE_FIREBASE_AI

CMakeLists.txt
```diff
@@ -345,8 +345,8 @@ if (FIREBASE_INCLUDE_STORAGE)
   list(APPEND DOCUMENTATION_ONLY_LIB_NAMES "firebase_storage_swig")
   list(APPEND PROJECT_LIST_HEADER "  X(Storage)")
 endif()
-if (FIREBASE_INCLUDE_VERTEXAI)
-  add_subdirectory(vertexai)
+if (FIREBASE_INCLUDE_FIREBASEAI)
+  add_subdirectory(firebaseai)
```
cynthiajoan, line 349: firebase_ai? I'm not quite sure about this one now.

cynthiajoan: After reading more code underneath, I think firebaseai is good.

CMakeLists.txt
a-maurice: It is also based on the directory structure. Since we decided on "firebaseai" without the underscore, I think it makes sense to not have it here either, to be consistent.

cynthiajoan: See some other cases which makes more sense to get FIREBASEAI all together while using upper cases. So let's keep it as FIREBASEAI

cynthiajoan (approved):

PR #1262, opened 2025-06-05
Analytics SetDefaultParameters

jonsimantov:
Add SetDefaultParameters API to Firebase Analytics Unity SDK.

This commit introduces the `FirebaseAnalytics.SetDefaultParameters()` method to the Unity SDK, wrapping the underlying C++ `firebase::analytics::SetDefaultEventParameters()` and `firebase::analytics::ClearDefaultEventParameters()` functions.

***

Calls to the new API have been integrated into the existing TestApp (UIHandler.cs) for manual verification and TestAppAutomated (UIHandlerAutomated.cs) for automated execution. These tests will exercise the API surface but not programmatically verify results from the Unity side.

***

### Type of Change
- [X] New feature. A non-breaking change which adds functionality.
***

### DIFF
--- a/analytics/src/FirebaseAnalytics.cs
+++ b/analytics/src/FirebaseAnalytics.cs
@@ -248,6 +248,30 @@ public static partial class FirebaseAnalytics {
     FirebaseAnalyticsInternal.SetConsentWithInts(consentSettingsMap);
   }
 
+  /// @brief Sets default event parameters.
+  ///
+  /// These parameters are logged with all events, in addition to the parameters passed to the
+  /// LogEvent() call. They are useful for logging common parameters with all events.
+  /// Default event parameters are overridden by event-specific parameters if the names are the
+  /// same. Default parameters are removed if the dictionary is null or empty.
+  ///
+  /// @param[in] parameters The dictionary of parameters to set.
+  ///   If null, clears all default parameters.
+  public static void SetDefaultParameters(
+      System.Collections.Generic.IDictionary<string, object> parameters) {
+    if (parameters == null || parameters.Count == 0) {
+      FirebaseAnalyticsInternal.ClearDefaultEventParameters();
+    } else {
+      StringList parameterNames = new StringList();
+      VariantList parameterValues = new VariantList();
+      foreach (var kvp in parameters) {
+        parameterNames.Add(kvp.Key);
+        parameterValues.Add(Firebase.Variant.FromObject(kvp.Value));
+      }
+      FirebaseAnalyticsInternal.SetDefaultEventParametersHelper(parameterNames, parameterValues);
+    }
+  }
+
   /// @brief Sets the duration of inactivity that terminates the current session.
   ///
   /// @note The default value is 30 minutes.
--- a/analytics/src/swig/analytics.i
+++ b/analytics/src/swig/analytics.i
@@ -87,6 +87,24 @@ void SetConsentWithInts(const std::map<int, int>& settings) {
   SetConsent(converted);
 }
 
+// Helper function to set default event parameters from vectors of strings and variants.
+void SetDefaultEventParametersHelper(
+    const std::vector<std::string>& parameter_names,
+    const std::vector<firebase::Variant>& parameter_values) {
+  if (parameter_names.size() != parameter_values.size()) {
+    firebase::LogError(
+        "SetDefaultEventParametersHelper given different list sizes (%d, %d)",
+        parameter_names.size(), parameter_values.size());
+    return;
+  }
+
+  std::map<std::string, firebase::Variant> default_parameters;
+  for (size_t i = 0; i < parameter_names.size(); ++i) {
+    default_parameters[parameter_names[i]] = parameter_values[i];
+  }
+  SetDefaultEventParameters(default_parameters);
+}
+
 }  // namespace analytics
 }  // namespace firebase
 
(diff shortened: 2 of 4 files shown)


analytics/src/FirebaseAnalytics.cs
```diff
@@ -15,11 +15,14 @@
  */
 
 using System.Threading.Tasks;
+using Firebase.Platform;
 
 namespace Firebase.Analytics {
 
 public static partial class FirebaseAnalytics {
 
+  private static readonly Firebase.Platform.ModuleLogger logger = new Firebase.Platform.ModuleLogger(typeof(FirebaseAnalytics).ToString());
```
jonsimantov: Let's hard code the tag instead of using typeof().ToString.

analytics/src/FirebaseAnalytics.cs
```diff
@@ -248,6 +251,59 @@ public static void SetConsent(System.Collections.Generic.IDictionary< ConsentTyp
     FirebaseAnalyticsInternal.SetConsentWithInts(consentSettingsMap);
   }
 
+  /// @brief Sets default event parameters.
+  ///
+  /// These parameters are logged with all events, in addition to the parameters passed to the
+  /// LogEvent() call. They are useful for logging common parameters with all events.
+  /// Default event parameters are overridden by event-specific parameters if the names are the
+  /// same. Default parameters are removed if the dictionary is null or empty.
```
jonsimantov, line 259: Let's change it so that default parameters are only cleared if the dictionary is null. *Individual* parameters will be removed (by the underlying implementation) if the value of a given key is null.

analytics/src/FirebaseAnalytics.cs
```diff
@@ -248,6 +251,59 @@ public static void SetConsent(System.Collections.Generic.IDictionary< ConsentTyp
...
+  ///
+  /// @param[in] parameters The dictionary of parameters to set.
+  ///   If null, clears all default parameters.
+  public static void SetDefaultEventParameters(
+      System.Collections.Generic.IDictionary<string, object> parameters) {
+    if (parameters == null || parameters.Count == 0) {
```
jonsimantov: Don't do this if count = 0, only if it's null.

analytics/src/FirebaseAnalytics.cs
```diff
@@ -248,6 +251,59 @@ public static void SetConsent(System.Collections.Generic.IDictionary< ConsentTyp
...
+              kvp.Key, e.ToString()));
+          // If adding to parameterNames succeeded but Variant.FromObject failed,
+          // we need to remove the key that was added optimistically.
+          if (parameterNames.Count > parameterValues.Count) {
+              parameterNames.RemoveAt(parameterNames.Count - 1);
+          }
+        }
+      }
+
+      if (parameterNames.Count == 0 && originalCount > 0) {
```
jonsimantov: I don't think we need this originalCount logic any more if we change it so an empty dictionary is just passed through verbatim.

analytics/src/FirebaseAnalytics.cs
```diff
@@ -248,6 +251,59 @@ public static void SetConsent(System.Collections.Generic.IDictionary< ConsentTyp
...
+      logger.LogMessage(LogLevel.Debug, "Cleared default event parameters.");
+    } else {
+      StringList parameterNames = new StringList();
+      VariantList parameterValues = new VariantList();
+      int originalCount = parameters.Count; // Store original count for later check
+
+      foreach (var kvp in parameters) {
+        try {
+          parameterNames.Add(kvp.Key);
+          parameterValues.Add(Firebase.Variant.FromObject(kvp.Value));
```
jonsimantov: Wouldn't it be cleaner to do the conversion first, and then add the name and value only once the conversion succeeds? This will prevent us from having to remove the value in the exception handler below.

analytics/src/swig/analytics.i
```diff
@@ -100,6 +118,9 @@ void SetConsentWithInts(const std::map<int, int>& settings) {
 %ignore firebase::analytics::LogEvent(const char*, const Parameter*, size_t);
 // Ignore SetConsent, in order to convert the types with our own function.
 %ignore firebase::analytics::SetConsent;
+// Ignore SetDefaultEventParameters and ClearDefaultEventParameters, as we handle them
```
jonsimantov: This comment is outdated as we don't ignore ClearDefaultEventParameters.

analytics/testapp/Assets/Firebase/Sample/Analytics/UIHandlerAutomated.cs
```diff
@@ -108,6 +109,34 @@ Task TestAnalyticsSetConsentDoesNotThrow() {
...
+      FirebaseAnalytics.SetDefaultEventParameters(new Dictionary<string, object>
+      {
+          { "auto_default_param_string", "auto_value_1" }
+      });
+
+      DebugLog("Automated Test: Setting multiple default parameters: {'auto_default_param_int', 789, 'auto_default_param_double', 12.34, 'auto_default_param_bool', false}");
+      FirebaseAnalytics.SetDefaultEventParameters(new Dictionary<string, object>
+      {
+          { "auto_default_param_int", 789 },
+          { "auto_default_param_double", 12.34 },
```
jonsimantov, line 126: We should also test with a value of null to make sure that that's passed through correctly.

analytics/src/FirebaseAnalytics.cs
jonsimantov: Also can we just confirm that Firebase.Variant.FromObject supports null properly, and will output a wrapped Variant::Null()?

PR #1271, opened 2025-06-16
Copy GA DLL for Windows Build and Play Mode if it exists

jonsimantov:
This change adds `analytics_win.dll` (if present) to the Windows build directory / Play Mode when the Firebase Analytics module is used.

The following modifications were made:

- Modified `editor/app/src/DllLocationPatcher.cs`:
    - Added logic to `OnPostProcessBuild` to copy `analytics_win.dll` from the project root to the build's plugin directory (`<project_name>_Data/Plugins/`) when building for StandaloneWindows or StandaloneWindows64, but only if the Firebase Analytics module is detected and the DLL exists.
- Created `editor/app/src/AnalyticsPlayModeSetup.cs`:
    - This new editor script subscribes to `EditorApplication.playModeStateChanged`.
    - When entering Play Mode on a Windows editor, it checks for the Firebase Analytics module and the existence of `analytics_win.dll` in the project root.
    - If both conditions are met, it copies `analytics_win.dll` to `Assets/Plugins/` to make it available during Play Mode.

These changes ensure that `analytics_win.dll` is correctly placed for both Windows builds and Play Mode sessions, facilitating proper Firebase Analytics functionality on the Windows platform.

***

***

### Type of Change
***


editor/app/src/AnalyticsPlayModeSetup.cs
```diff
@@ -0,0 +1,49 @@
+using System;
```
a-maurice, line 1: This file needs to be added to the editor/app/CMakeLists.txt

editor/app/src/AnalyticsPlayModeSetup.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (Application.platform == RuntimePlatform.WindowsEditor) {
+                Type firebaseAnalyticsType = Type.GetType("Firebase.Analytics.FirebaseAnalytics, Firebase.Analytics");
+                if (firebaseAnalyticsType != null) {
+                    if (File.Exists("./analytics_win.dll")) {
+                        string fullSourcePath = Path.GetFullPath("./analytics_win.dll");
+                        // string destinationDirectory = "./"; // This is DestinationDir
+                        string fullDestinationPath = Path.GetFullPath(Path.Combine(DestinationDir, "analytics_win.dll"));
+
+                        if (fullSourcePath.Equals(fullDestinationPath, StringComparison.OrdinalIgnoreCase)) {
+                            Debug.Log("Firebase Analytics: analytics_win.dll is already in the project root. No copy needed for Play Mode.");
```
a-maurice, line 27: We probably don't want to log if nothing is being done, cause that'll be pretty spammy.

editor/app/src/AnalyticsPlayModeSetup.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    private static void OnPlayModeStateChanged(PlayModeStateChange state) {
+        if (state == PlayModeStateChange.EnteredPlayMode) {
+            if (Application.platform == RuntimePlatform.WindowsEditor) {
+                Type firebaseAnalyticsType = Type.GetType("Firebase.Analytics.FirebaseAnalytics, Firebase.Analytics");
+                if (firebaseAnalyticsType != null) {
+                    if (File.Exists("./analytics_win.dll")) {
+                        string fullSourcePath = Path.GetFullPath("./analytics_win.dll");
+                        // string destinationDirectory = "./"; // This is DestinationDir
+                        string fullDestinationPath = Path.GetFullPath(Path.Combine(DestinationDir, "analytics_win.dll"));
```
a-maurice, line 24: Nothing ever changes DestinationDir afaik, so this will always be ./analytics_win.dll, which will be the same as the fullSourcePath above. I'm not clear what it is trying to do here.

PR #1317, opened 2025-08-21
[Firebase AI] Add Thought summaries

a-maurice:
Add logic for receiving thought summaries when generating content.
***

Running tests locally.
***

### Type of Change
***

### DIFF
--- a/firebaseai/src/GenerateContentResponse.cs
+++ b/firebaseai/src/GenerateContentResponse.cs
@@ -53,10 +53,26 @@ public readonly struct GenerateContentResponse {
   /// </summary>
   public string Text {
     get {
-      // Concatenate all of the text parts from the first candidate.
+      // Concatenate all of the text parts that aren't thoughts from the first candidate.
       return string.Join(" ",
           Candidates.FirstOrDefault().Content.Parts
-          .OfType<ModelContent.TextPart>().Select(tp => tp.Text));
+          .OfType<ModelContent.TextPart>().Where(tp => !tp.IsThought).Select(tp => tp.Text));
+    }
+  }
+  
+  /// <summary>
+  /// A summary of the model's thinking process, if available.
+  /// 
+  /// Note that Thought Summaries are only available when `IncludeThoughts` is enabled
+  /// in the `ThinkingConfig`. For more information, see the
+  /// [Thinking](https://firebase.google.com/docs/ai-logic/thinking) documentation.
+  /// </summary>
+  public string ThoughtSummary {
+    get {
+      // Concatenate all of the text parts that are thoughts from the first candidate.
+      return string.Join(" ",
+          Candidates.FirstOrDefault().Content.Parts
+          .OfType<ModelContent.TextPart>().Where(tp => tp.IsThought).Select(tp => tp.Text));
     }
   }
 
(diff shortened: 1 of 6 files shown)


firebaseai/src/GenerateContentResponse.cs
```diff
@@ -65,7 +81,8 @@ public string Text {
   /// </summary>
   public IReadOnlyList<ModelContent.FunctionCallPart> FunctionCalls {
     get {
-      return Candidates.FirstOrDefault().Content.Parts.OfType<ModelContent.FunctionCallPart>().ToList();
+      return Candidates.FirstOrDefault().Content.Parts
+          .OfType<ModelContent.FunctionCallPart>().Where(tp => tp.IsThought).ToList();
```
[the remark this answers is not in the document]
a-maurice: Fixed

firebaseai/src/ModelContent.cs
```diff
@@ -337,20 +405,25 @@ internal static ModelContent FromJson(Dictionary<string, object> jsonDict) {
       jsonDict.ParseObjectList("parts", PartFromJson, JsonParseOptions.ThrowEverything).Where(p => p is not null));
   }
 
-  private static InlineDataPart InlineDataPartFromJson(Dictionary<string, object> jsonDict) {
+  private static InlineDataPart InlineDataPartFromJson(Dictionary<string, object> jsonDict,
+      bool? isThought, string thoughtSignature) {
     return new InlineDataPart(
       jsonDict.ParseValue<string>("mimeType", JsonParseOptions.ThrowEverything),
       Convert.FromBase64String(jsonDict.ParseValue<string>("data", JsonParseOptions.ThrowEverything)));
```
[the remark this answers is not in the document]
a-maurice: Fixed

firebaseai/src/GenerateContentResponse.cs
```diff
@@ -65,7 +81,8 @@ public string Text {
   /// </summary>
   public IReadOnlyList<ModelContent.FunctionCallPart> FunctionCalls {
```
cynthiajoan, line 82: Do you also need to add the same logic for getting InlineDataPart?

a-maurice: There currently isn't a helper function for that. I guess I could add one...

firebaseai/src/Internal/InternalHelpers.cs
```diff
@@ -234,6 +234,20 @@ public static ModelContent ConvertToModel(this ModelContent content) {
...
+      jsonDict.Add(key, value.Value);
+    }
+  }
+  
+  public static void AddIfHasValue<T>(this JsonDict jsonDict, string key,
+      T value) where T : class {
+    if (value != null) {
+      jsonDict.Add(key, value);
+    }
+  }
```
cynthiajoan, line 250: why we would need to functions, the only difference I can find is the ? after T in parameters

firebaseai/src/GenerateContentResponse.cs
a-maurice: If I do this, I'll do it in a separate PR

firebaseai/src/Internal/InternalHelpers.cs
a-maurice: By default, struct's are not-nullable, which is why I need the version with T?, and retrieve the underlying value with .Value.  Classes can be null, and by default doing T? with a class will result in a warning. So doing two methods helps prevent that warning.

PR #1343, opened 2025-10-02
feat: add ImageConfig and new FinishReason values

thatfiredev:
Adds new `FinishReason` values to handle image-specific cases. Introduces `ImageConfig` to allow specifying image generation parameters like aspect ratio, and integrates it into the `GenerationConfig` struct.


paulb777:
Continued in #1422

PR #1385, opened 2026-01-30
Feat: add in limited use tokens

AustinBenoit:
Feat: add in limited use tokens

This changes the current interface implementation `IAppCheckProvider` to have both a 
***

I have the new limited use token locally 

Android: 

image

iOS:
image

Desktop:
image

Along with running the integration tests
***

### Type of Change
- [X] New feature. A non-breaking change which adds functionality.
***

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -122,7 +122,6 @@ Release Notes
     - Firebase AI: Add support for `CachedContentTokenCount` and `CacheTokensDetails`
       in `UsageMetadata`.
 
-
 ### 13.7.0
 -   Changes
     - General: Update to Firebase C++ SDK version 13.4.0.


app_check/src/FirebaseAppCheck.cs
```diff
@@ -128,6 +133,24 @@ public System.Threading.Tasks.Task<AppCheckToken>
...
+  public System.Threading.Tasks.Task<AppCheckToken>
+      GetLimitedUseAppCheckTokenAsync() {
+    ThrowIfNull();
+    return appCheckInternal.GetLimitedUseAppCheckTokenAsync().ContinueWith(task => {
+      if (task.IsFaulted) {
+        throw task.Exception;
+      }
+      AppCheckTokenInternal tokenInternal = task.Result;
+      return AppCheckToken.FromAppCheckTokenInternal(tokenInternal);
+    });
```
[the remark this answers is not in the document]
AustinBenoit: Umm yeah I'll come back to this and clean both limiteduse and get token together.

app_check/src/FirebaseAppCheck.cs
```diff
@@ -96,14 +99,16 @@ public static void SetAppCheckProviderFactory(IAppCheckProviderFactory factory)
     appCheckFactory = factory;
     // Clear out the Providers that were previously made. When future calls to
     // GetToken fails to find a provider in the map, it will use the new factory
-    // to create a new provider.
+     // to create a new provider.
```
a-maurice: nit: extra space

app_check/src/swig/app_check.i
```diff
@@ -177,6 +215,19 @@ void SetGetTokenCallback(GetTokenFromCSharp get_token_callback) {
   }
 }
 
+// Called from C# to register the C# to call to get a limited use token.
+void SetGetLimitedUseTokenCallback(GetTokenFromCSharp get_token_callback) {
```
a-maurice: I feel like we should merge this with SetGetTokenCallback.  I don't think there is a scenario when we actually want to set them separately, and having them set separately is a bit confusing because they are both calling into SetAppCheckProviderFactory and doing all the setup/teardown.

app_check/src/FirebaseAppCheck.cs
AustinBenoit: Done

app_check/src/swig/app_check.i
AustinBenoit: Ah that does make more sense. Done!

a-maurice (approved):

PR #1388, opened 2026-02-04
Add in in test retry for AI.

AustinBenoit:
The github runners will retry but at whole sdk level. This means even if we encounter a retryable error in one test we must retry the whole product. This means we wasting our credits. In particular for our limited image generation ones. We can limit this by retrying in the test itself. The only con here is potenitaly retrying something that is doomed to fail. I have tried to put in some smarts to prevent this.

In the tests we can see a 429 resource exhausted that come in two flavours.

"code": 429,
    "message": "Resource exhausted. Please try again later. Please refer to https://cloud.google.com/vertex-ai/generative-ai/docs/error-code-429 for more details.",

and

"code": 429,
    "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits. To monitor your current usage, head to: https://ai.dev/rate-limit. \n* Quota exceeded for metric: generativelanguage.googleapis.com/predict_requests_per_model_per_day_paid_tier_1, limit: 70, model: imagen-4.0-generate",

The exponential backoff should be helpful for  "Resource exhausted. Please try again later. " but no the quota exceded one.

***

Tested this locally and saw some good recoveries that prevented the whole test run failing.

***

### Type of Change
***

### DIFF
--- a/firebaseai/src/Internal/HttpHelpers.cs
+++ b/firebaseai/src/Internal/HttpHelpers.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Net.Http;
+using System.Net;
 using System.Threading.Tasks;
 
 namespace Firebase.AI.Internal
@@ -106,12 +107,27 @@ namespace Firebase.AI.Internal
       }
 
       // Construct the exception with as much information as possible.
-      throw new HttpRequestException(
+      var ex = new HttpRequestException(
         $"HTTP request failed with status code: {(int)response.StatusCode} ({response.ReasonPhrase}).\n" +
         $"Error Content: {errorContent}",
-        null,
-        response.StatusCode
+        null
       );
+      ex.Data["StatusCode"] = response.StatusCode;
+
+      throw ex;
+    }
+  }
+
+  // Extension to get the StatusCode from the exception.
+  internal static class HttpRequestExceptionExtensions
+  {
+    internal static HttpStatusCode? GetStatusCode(this HttpRequestException exception)
+    {
+      if (exception.Data.Contains("StatusCode"))
+      {
+        return (HttpStatusCode)exception.Data["StatusCode"];
+      }
+      return null;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


firebaseai/src/Internal/HttpHelpers.cs
```diff
@@ -110,6 +110,7 @@ internal static async Task ValidateHttpResponse(HttpResponseMessage response)
         $"HTTP request failed with status code: {(int)response.StatusCode} ({response.ReasonPhrase}).\n" +
         $"Error Content: {errorContent}"
       );
+      ex.Data["StatusCode"] = (int)response.StatusCode;
```
a-maurice: Instead of trying to pass it along via this, it might be better to use one of the other constructors that takes in a StatusCode: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequestexception.-ctor?view=net-10.0#system-net-http-httprequestexception-ctor(system-string-system-exception-system-nullable((system-net-httpstatuscode)))

Then below you can use the built in StatusCode property of the exception.

AustinBenoit: I tired to add in the newer constructor but we currently have  Unity 2021.3.43f1 with .NET Standard 2.1 and that constructor needs .NET 5+ so I could not make this work.

firebaseai/src/Internal/HttpHelpers.cs
```diff
@@ -108,10 +109,25 @@ internal static async Task ValidateHttpResponse(HttpResponseMessage response)
...
+      ex.Data["StatusCode"] = response.StatusCode;
 
       throw ex;
     }
   }
+
+  // Extension to get the StatusCode from the exception.
+  internal static class HttpRequestExceptionExtensions
+  {
+    internal static HttpStatusCode? GetStatusCode(this HttpRequestException exception)
```
a-maurice, line 124: Could you add a comment explaining that Unity uses an older dotNet standard that doesn't include the StatusCode property on the HttpRequestException?

PR #1393, opened 2026-02-17
Chore: move the https and interop files to app so multiple sdks can use them

AustinBenoit:
Chore: move the https and interop files to app so multiple sdks can use them
***

Ran local test with firebase AI and integration test should still pass.
***

### Type of Change
***


app/src/internal/FirebaseInterops.cs
```diff
@@ -20,7 +20,7 @@
 using System.Reflection;
 using System.Threading.Tasks;
 
-namespace Firebase.AI.Internal
+namespace Firebase
```
a-maurice: This and the other file should be in Firebase.Internal

cmake/unity_mono.cmake
```diff
@@ -270,6 +275,7 @@ macro(mono_add_internal name output_type)
   set(VAR_ROOT_NAMESPACE ${UNITY_MONO_MODULE})
   set(VAR_ASSEMBLY_NAME ${UNITY_MONO_ASSEMBLY_NAME})
   set(VAR_FRAMEWORK_VERSION ${UNITY_MONO_FRAMEWORK_VERSION})
+  set(VAR_LANG_VERSION ${UNITY_MONO_LANG_VERSION})
```
a-maurice, line 278: Just wondering, what inspired this change?

AustinBenoit: Ah, in app/src/internal/FirebaseInterops.cs there is some C# 9.0+ only  features

```
private static readonly Lazy<string> _sdkVersionFetcher = new(() =>

if (taskObject is not Task appCheckTokenTask)
// ...
if (taskObject is not Task tokenTask)

```
Per https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-9

So when these are bundle as src the Unity editor compiles them as c# 9.0+. Depending on the supported version of the editor. 
However when we compile them as src in cmake/unity_mono.cmake and end up calling roslyn directly it seems to default to the oldest support version which in C# 8 (I think it may be an older version). The compiler supports a bunch of version so I just picked the oldest C# lang version that supported the existing functionality.

app/src/internal/FirebaseInterops.cs
AustinBenoit: Done.

a-maurice (commented):
For the new source files, you should include the .meta files along with them.  Generally you can just copy them from your Unity project, but that way they will have consistent guids associated with them across releases.

AustinBenoit:
> For the new source files, you should include the .meta files along with them. Generally you can just copy them from your Unity project, but that way they will have consistent guids associated with them across releases.

@a-maurice Done

app/src/internal/FirebaseInterops.cs.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
+guid: bc30602676ea492eab91b6ff7cd6ba3b
+labels:
```
a-maurice: Actually, can you remove the labels from here. Same with the other meta file.

AustinBenoit: Done

a-maurice (approved):

AustinBenoit:
Test technically pass just failing with an out of quota exception. Other platforms work so I'll say this is passed. https://github.com/firebase/firebase-unity-sdk/actions/runs/22503906490/job/65208864451

PR #1429, opened 2026-04-16
[App] Add link.xml to preserve Auth types for reflection

AustinBenoit:
Add a `link.xml` file to the core `App` module in `app/src/internal/`. This file prevents Unity's IL2CPP compiler from aggressively stripping `Firebase.Auth` types (`FirebaseAuth` and `FirebaseUser`) when they are accessed via reflection by other modules like `Functions` or `Firebase AI` (via `FirebaseInterops.cs`).
This resolves runtime `INTERNAL` errors on iOS/tvOS where the `Functions` SDK failed to retrieve Auth tokens due to missing symbols, resulting in backend assertion failures.
***
*   Verified that the `tokenTest` in the Functions integration test app passes on the iOS simulator after adding `link.xml`.
*   Verified that `link.xml` is correctly bundled in both `.unitypackage` and UPM (`.tgz`) generated artifacts via the build scripts.
*  Full run of tests verifying that they all pass
***
### Type of Change
***

### DIFF
--- a/app/src/internal/link.xml.meta
+++ b/app/src/internal/link.xml.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: 178c4b1924574f9c95f7a7f28adfd218
+labels:
+- gvh
+- gvh_version-13.10.0
+- gvhp_exportpath-Firebase/FirebaseApp/Internal/link.xml
+timeCreated: 1480838400
+DefaultImporter:
+  userData:
+  assetBundleName:
+  assetBundleVariant:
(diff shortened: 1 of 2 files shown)


a-maurice (approved):

app/src/internal/link.xml.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
+guid: 178c4b1924574f9c95f7a7f28adfd218
+labels:
```
a-maurice: Remove these labels, we don't want the labels there by default

AustinBenoit: Done

a-maurice (approved):

PR #1452, opened 2026-05-07
Fix iOS simulator test runs: Let Xcode decide the best architecture

AustinBenoit:
Fix iOS simulator test runs: Build both x86 and arm. 
***

Run the integration tests.  
***

### Type of Change
***

### DIFF
--- a/scripts/build_scripts/build_zips.py
+++ b/scripts/build_scripts/build_zips.py
@@ -79,6 +79,7 @@ ANDROID_SUPPORT_ARCHITECTURE = ["armeabi-v7a", "arm64-v8a", "x86", "x86_64"]
 MACOS_SUPPORT_ARCHITECTURE = ["x86_64", "arm64"]
 
 g_target_architectures = []
+g_target_devices = []
 g_cpp_sdk_realpath = ""
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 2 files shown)


scripts/build_scripts/build_zips.py
```diff
@@ -816,12 +935,16 @@ def main(argv):
...
+    make_ios_multi_arch_build(cmake_setup_args)
   else:
     subprocess.call(cmake_setup_args)
     if (not FLAGS.gen_swig_only):
-      if is_windows_build() or is_ios_build():
+      if is_windows_build():
         # no make command in windows or when using Xcode generator. TODO make config passable
         subprocess.call(["cmake", "--build", ".", "--config", "Release"])
+      elif is_ios_build(): 
+        subprocess.call(["cmake", "--build", ".", "--config", "Release", "--", "-arch", "arm64", "-arch", "x86_64"])
```
a-maurice: Presumably this can't happen, because it is already checking for is_ios_build above now?

a-maurice (approved):

scripts/build_scripts/build_zips.py
AustinBenoit: Ahh yeah. I have removed it and now removed it from the following section too.

PR #1517, opened 2026-09-03
Remove testapp's package-lock file

a-maurice:
Remove the testapp's package-lock file from the Firebase Functions. The important part for users is the package.json file, the package-lock causes confusion about what is required, and would need too many regular updates.
***

***

### Type of Change
***


paulb777 (approved):
For SDK testing, we want to float forward from required minimums so that we detect problems with future versions ASAP.

We don't want exact reproducibility like an app would

Issue #1519, opened 2026-09-11
[FR]: iOS: the simulator slice of the Unity C++ libraries is x86_64-only — please add arm64-simulator


TimRobinNet:
### Description

Hello dear Firebase team,

the iOS C++ libraries shipped with the Unity SDK (libFirebaseCppApp.a, libFirebaseCppAnalytics.a, libFirebaseCppCrashlytics.a, libFirebaseCppMessaging.a) do contain a simulator slice — but only for x86_64. There is no arm64-simulator code, so a Unity project using Firebase cannot be built for the iOS Simulator on Apple Silicon.

Verified against Firebase Unity SDK 13.16.0 (latest at time of writing):

```
$ lipo -info libFirebaseCppApp.a
Architectures in the fat file: ... are: x86_64 arm64
```

```
$ otool -arch arm64 -l libFirebaseCppApp.a | grep -A4 LC_BUILD_VERSION
  33  platform 2      minos 12.0   sdk 26.2      # PLATFORM_IOS — device
```

```
$ otool -arch x86_64 -l libFirebaseCppApp.a | grep -A4 LC_BUILD_VERSION
  33  platform 7      minos 12.0   sdk 26.2      # PLATFORM_IOSSIMULATOR
```

So the build intentionally produces a device slice and a simulator slice; the simulator one is just still x86_64, which was sufficient before Apple Silicon.

Result when building for an arm64 simulator (Xcode 27.0, target arm64-apple-ios17.0-simulator):

```
ld: building for 'iOS-simulator', but linking in object file
(libFirebaseCppApp.a[arm64][2](log.cc.o)) built for 'iOS'
```

Note this cannot be fixed within the current packaging: a fat archive is keyed by CPU type, so it cannot hold both an arm64-device and an arm64-simulator slice. Shipping arm64-simulator requires an XCFramework (or parallel device/simulator library sets).

The C++ SDK already builds exactly what is needed — [firebase-cpp-sdk/build_scripts/ios/build.sh](https://github.com/firebase/firebase-cpp-sdk/blob/main/build_scripts/ios/build.sh):

```
readonly SUPPORTED_PLATFORMS=(device simulator)
readonly SIMULATOR_ARCHITECTURES=(arm64 x86_64)
...
sysroot_arg="-DCMAKE_OSX_SYSROOT=iphonesimulator"
```

[...]

PR #1520, opened 2026-09-14
build(ios): allow building an arm64 simulator slice

chagretes:
### Description

Makes `--device=simulator --architecture=arm64` produce what it advertises, so an
arm64 iOS simulator slice can be built. Relates to #1519.

Two things currently prevent it:

**1. `cmake/unity_ios.cmake` discards the requested architecture.**
`build_zips.py` computes `-DCMAKE_OSX_ARCHITECTURES` from `--architecture` and
passes it, but the toolchain then sets the same variable unconditionally with
`CACHE`, so the flag never survives. It is now applied only as a default.

**2. `build_zips.py`'s iOS loop ignores `--architecture`.**
It iterates `IOS_CONFIG_DICT[device]["architecture"]` rather than the flag, then
skips arm64 simulator unconditionally. That skip exists for a real reason — an
arm64 simulator slice cannot coexist with an arm64 device slice inside one fat
static library — but it only applies when device and simulator are built
together. Building the simulator alone has nothing to collide with, so the skip
is now conditional on `len(g_target_devices) > 1`.

Both changes are build-only; no library or C# code is touched. Default behaviour
is unchanged: with no `--architecture`, and with device and simulator built
together, the same architectures are produced as before.

### Testing

```
python3 scripts/build_scripts/build_zips.py --platform=ios \
    --targets=analytics --targets=crashlytics \
    --device=simulator --architecture=arm64 --unity_root=<unity>
```

Produces `libFirebaseCppApp.a`, `libFirebaseCppAnalytics.a` and
`libFirebaseCppCrashlytics.a` as arm64 with `LC_BUILD_VERSION platform 7`
(IOSSIMULATOR) rather than platform 2 (device):

```
$ lipo -info libFirebaseCppAnalytics.a
Non-fat file: ... is architecture: arm64

$ ar x libFirebaseCppAnalytics.a && otool -l analytics_common.o | grep -A2 LC_BUILD_VERSION
      cmd LC_BUILD_VERSION
  platform 7
    minos 14.0
```

Before the change the same command yields x86_64 (the flag is ignored and the
toolchain default wins).

Those libraries then linked into an arm64 iOS simulator build of a large Unity
game (Unity 6000.3, IL2CPP, ~50 CocoaPods), which installed on an iOS 26.5
simulator, launched, and exercised Firebase-dependent startup paths.

I also confirmed the default path is unaffected by building with no
`--architecture` flag and seeing the same architectures as before the change.

There is no automated test suite covering these build scripts, so the
verification above is manual — happy to add something if you have a preferred
place for it.

### Build prerequisites

For anyone reproducing locally, beyond the Firebase C++ prerequisites: Mono
(the CMake step wants `mono` plus `msbuild` or `xbuild`), Swig 4+, a Unity
install with iOS Build Support passed via `--unity_root`, and `absl-py`
importable by whatever `python` the sub-scripts resolve to.

### Note on scope

This does not attempt to solve distribution. Shipping device and simulator
slices together still needs an XCFramework (or a parallel simulator artifact),
which is the larger piece discussed in #1519. This only makes the library
buildable, which that work would depend on anyway.

### Type of Change


chagretes:
Thanks — both applied in 285edeb.

**Explicit device check:** agreed, `len(g_target_devices) > 1` was relying on `SUPPORT_DEVICE` only ever having two entries. Now checks for `"device"` and `"simulator"` by name.

**Quoting in the CMake guard:** applied, since quoting is the safer form and costs nothing. One correction for the record, though: the failure described does not reproduce. `if(VAR STREQUAL "")` treats `VAR` as a variable *name*, so it is not expanded into multiple arguments — that would need an explicit `${VAR}` dereference. Both forms behave identically:

| `CMAKE_OSX_ARCHITECTURES` | unquoted | quoted |
|---|---|---|
| (undefined) | `arm64;x86_64` | `arm64;x86_64` |
| `""` | `""` | `""` |
| `arm64` | `arm64` | `arm64` |
| `arm64;x86_64` | `arm64;x86_64` | `arm64;x86_64` |

No syntax or runtime error in any case. Flagging it so the reasoning here is not taken as a fixed bug.

**Re-verified after both changes:**

- `--device=simulator --architecture=arm64` still produces an arm64 library with `LC_BUILD_VERSION platform 7` (IOSSIMULATOR).
- The default path is untouched: with both devices and no `--architecture`, selection is still `device/arm64` + `simulator/x86_64`, exactly as before this PR. I checked this against the patched source across six device/flag combinations, since that path is the one that ships.

a-maurice (approved):
Thanks for sending this in, looks good. I'll follow up in the other issue about other plans to address this for the release build, but hopefully this can unblock some folks in the meantime.

